PR #23, opened 2025-09-17
Pull gir lib path from pkgconf

fllmn:
Had some issues getting it running on Fedora since `gobject-introspection` location seems to be distribution dependent. I don't know pixi well enough to say if this is the correct way of doing it or not.


knipknap:
Awesome, that is a better way to do it, thanks!

PR #36, opened 2025-09-27
Update serial.py to include higher baud rates

loneregister:
Update serial.py to include higher baud rates. Specifically, the baud rate of 1,000,000 for Monport Mega 70w laser which uses this baud rate.


knipknap:
Awesome, Thank you!

PR #39, opened 2025-09-30
Resolve flatpak linting errors

fllmn:
**Remove Icon from appstream:**
There doesn't seem to be a way to set icon in the appstream without the linter complaining, removing it and instead adding a "post-install" phase seems to work.

**Change branch to build from main:**
There's a bit of causality problem, the appstream is currently broken so in order to get a working appstream when building the image the branch needs to point to main.. which will be this commit once it merges. Once a new release tag is available we can move to tracking that.

**Add pyclipper dependency**

**Minor cleanups**

### DIFF
--- a/flatpak/com.barebaric.rayforge.yml
+++ b/flatpak/com.barebaric.rayforge.yml
@@ -26,6 +26,9 @@ modules:
   - name: rayforge
     buildsystem: simple
     build-commands:
+      - install -Dm644 data/${FLATPAK_ID}.svg /app/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg
+      - install -Dm644 data/${FLATPAK_ID}.desktop /app/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm644 data/${FLATPAK_ID}.metainfo.xml /app/share/appdata/${FLATPAK_ID}.metainfo.xml
       - pip3 install --prefix=/app --no-deps --no-build-isolation .
     sources:
       - type: git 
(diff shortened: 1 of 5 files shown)


knipknap:
I suspect that removing the icon line from the appstream breaks the icon for the Snap package. I'll have a closer look soon.

fllmn:
Alright, then let me just revert it and see if the linting errors is acceptable. From having looked through a lot of other Flathub submissions barely anyone is compliant. It doesn't help that there is basically no guides or proper documentation

knipknap:
I am still testing, the Snap is still building since half an hour :smile:. Maybe somehow it is not a problem. Snap also has some very strange issues and basically no help - story of packaging anything ever, really :-/. Windows packaging is even worse.

knipknap:
Ok, so the Snap built now (with the commit before you reverted) and it indeed fails.

> Icon 'com.barebaric.rayforge' specified in desktop file 'share/applications/com.barebaric.rayforge.desktop' not found in prime directory.

But I *think* a workaround could be found for the Snap if you get stuck with the Flatpak. Let me know if you need help, I can have a look.

fllmn:
Alright, thanks for checking! I've reverted back the icon line. No more changes planned.

flatpak/com.barebaric.rayforge.yml
```diff
@@ -11,7 +11,6 @@ finish-args:
   - --device=all
   - --env=DISPLAY=:0
   - --share=network
-  - --talk-name=org.freedesktop.DBus
```
fllmn, line 14: Oh yes this as well, I know we talked about this on a previous PR. I've not found anything that breaks by removing it. It seems like exemptions from the linting errors are merged into a separate repository so I'd prefer starting without it

knipknap: Ok, since I also can't remember let's try leaving it out! :smile:

Issue #34, opened 2025-09-26
Implement a package management for plugins and other data sharing


knipknap:
Rayforge needs a robust, extensible plugin system that allows community contributions and easy management. Below is a draft architecture proposal for plugin support.

---

# Rayforge Plugin Architecture

### 1. Plugin Discovery & Loading (via pluggy)

```python
import pluggy

hookspec = pluggy.HookspecMarker("rayforge")
hookimpl = pluggy.HookimplMarker("rayforge")

class RayforgeSpec:
    """Core hook specifications"""
    
    @hookspec
    def rayforge_init(self, config):
        """Called when plugin is initialized"""
    
    @hookspec
    def before_generate(self, params):
        """Called before object generation"""
    
    @hookspec
    def after_generate(self, result):
        """Called after object generation"""
```

### 2. Plugin Manager

```python
from pathlib import Path
import yaml
import git

class PluginManager:
    def __init__(self):
        self.config_dir = Path("~/.config/rayforge/plugins").expanduser()
        self.config_dir.mkdir(parents=True, exist_ok=True)
        self.pm = pluggy.PluginManager("rayforge")
        self.pm.add_hookspecs(RayforgeSpec)

    def get_plugin_metadata(self, plugin_dir):
        metadata_file = plugin_dir / "rayforge_plugin.yaml"
        if metadata_file.exists():
            return yaml.safe_load(metadata_file.read_text())
        return None

    async def install_plugin(self, repo_url):
        """Install plugin from GitHub"""
        plugin_name = repo_url.split("/")[-1]
        plugin_dir = self.config_dir / plugin_name
        
        # Clone/update repo
        if plugin_dir.exists():
            repo = git.Repo(plugin_dir)
            repo.remotes.origin.pull()
        else:
            git.Repo.clone_from(repo_url, plugin_dir)
        
        # Load dependencies
        deps = self.get_plugin_metadata(plugin_dir).get("dependencies", [])
        for dep in deps:
            await self.install_plugin(dep)
        
        # Register plugin
        self.pm.register(importlib.import_module(f"{plugin_name}.plugin"))

    def uninstall_plugin(self, plugin_name):
        """Remove plugin"""
        plugin_dir = self.config_dir / plugin_name
        if plugin_dir.exists():
            shutil.rmtree(plugin_dir)
```

### 3. Plugin Metadata (plugin.yaml)

```yaml
name: my-plugin
version: "1.0.0"
description: "A cool rayforge plugin"
[...]

PR #40, opened 2025-10-02
Add material test grid tool and execution preview 

heathkh:
This adds two new features and some documentation for developers

1.) A material test grid generator tool useful for tuning the laser cutter settings for a given material
Screenshot from 2025-10-02 00-26-22

The design of this feature is described in docs/design/material_test_generator.md file

2.) Extends the "2D" and "3D" view modes with a new "Execution Preview" view mode.  This shows the execution of the cut with a video-player like interface.    
Screenshot from 2025-10-02 00-24-14

The design of this feature is described in docs/design/preview_playback.md 

3.)  I added a new docs/architecture.md file that is a high-level summary of the software design useful for me orienting with the codebase.


knipknap:
Oh wow, that is a huge feature, and nicely done, too! I'd be glad to include it as soon as possible.
Especially love the live preview, that was something I wanted to have for a while already!

To get this merged, I found a couple of minor issues that need addressing. I also have some higher-level thoughts on the architecture that I'd love to discuss with you.

## Required Changes (Blockers for Merge)

In testing, I found only some minor issues:

- In the step settings, the "preset" dropdown is visually different from the "engrave" dropdown. Could we use the Adw popover for consistency throughout the app?
- The mainwindow "3D" toggle button no longer works.

## Architectural Discussion Points

Overall, the architecture and UI look great and are very much how I would have done it :-). The producer, step settings, and renderer are excellent. I do have a few thoughts on the general approach that I'd like to get your opinion on. We can tackle these now or in follow-up PRs.

- I wonder if a dedicated layer type is really warranted. I think it makes the feature a little less discoverable - couldn't we just treat the material test step like any other step, allowing to mix and match it with other operations in the same layer? This would also remove the need for a new core object.
- I am similarly wondering if the view couldn't be integrated in the main existing 2D view. Even in my short testing I found myself trying edit and resize workpieces while in the "preview view". I think users might prefer a "simulate" toggle that displays the buttons, crosshair, and heat map within the main view.
- The simulator should probably also open the recently added gcode preview and highlight the current gcode row.
- Ops rendering encapsulation: I thought about extending the existing [cairoencoder](https://github.com/barebaric/rayforge/blob/main/rayforge/pipeline/encoder/cairoencoder.py) by an "OpsColorConfig" to allow the sort of color coding that you new implemented in the previewoverlay. That object could encapsulate colors that right now are fetched from the theme in a couple of locations (2dview, 3dview). My main concern is about the maintenance of Ops consumers - this has proven to be the largest slow down for me in development, because every change in the Ops objects requires downstream updates in producers, transformers, 2dview, 3dview, and now potentially the preview. Just to explain why I am thinking hard about encapsulating as much of this logic as possible.

## Future considerations (for context)

- The next task on my TODO list is "Recipe support". The idea behind that is creating savable workplans as a catalog of "recipes". Recipes would allow to define the context (i.e. Machine, Material, etc.) so they could be manually or automatically assigned when editing workpieces. This feature has other prerequisites, such as allowing to assign a material to each workpiece, rendering materials in different colors, ... I hadn't implemented it yet precisely because I didn't have a clear picture of how material testing would fit in. Your work here is a critical piece of that puzzle, though I still don't have a clear direction on how to best integrate it all. :smile:
- Addon support. Long term, I'm trying to design Rayforge with a small core, organizing features into packages (#34). I see material test generators living in such packages. Materials and machine profiles including recipes, too. This is another reason it's good to encapsulate the feature into dedicated Python modules - to make moving it easier later.

All that said, this is a fantastic implementation. My main concerns before merging are just fixing the two minor issues and getting your thoughts on the dedicated layer type. We can definitely make additional architectural changes after the initial merge if you prefer.

knipknap:
Dropping a reference to more docs for consideration (#41, #42)

heathkh:
Thanks for the very thoughtful notes, I really appreciate getting the big-picture roadmap.  I can tell you have managed long lived software projects before.  The thought you put into maintainability and design give me great confidence in the potential long-term impact of this tool. 

Also, thanks for reading this PR, it is way bigger than I consider good practice.  In the spirit of rapid-prototyping and getting early feedback / guidance I wanted to get it to you sooner.

Let me take a stab at fixing the blockers and try out the cleaner design suggestions you made.

I think the test-grid-generator would be a good fit for a plugin.  It could be a useful "field test" of the plugin architecture to try to migrate the material test grid generator feature to the plugin system.  I'll scan the design doc for the plugin system. 

Question: Would you prefer to merge a PR with the test-grid-generator feature baked in and then later move it to a plugin, or ship it now and then move it to a plugin as a use-case when fleshing out the plugin system?

knipknap:
Absolutely no rush to move it to a plugin, in fact I think it would be premature. The plugin API is a draft at best with not even a prototype in place.

Thanks a lot for the work and thought you put into this!

heathkh:
## I've made the following changes

- Removed the material test grid layer and use standard layer instead
- In the material test grid step settings, the "preset" dropdown use the Adw popover for consistency
image

- Standardize docs and code on the term "Simulation" rather than "execution preview / playback"
- Fix mainwindow "3D" toggle button 
- Added a "Simultation" toolbar toggle button ("play icon")
- Reverted to simpler design where you can toggle on "3D Mode" and "Simulation Mode" independently rather than treating 2D, 3D, Simulation as 3 different views.  Right now the simulation and 3d mode are mutually exclusive, but that constraint could probably be removed as part of unifying the cairos ops renderer
image

- The simulator mode opens the new gcode preview and highlight the current gcode row in sync with simulation playback
image

## Remaining for future work:
- Ops rendering encapsulation

heathkh:
When you have time, I'm ready for another round of review...

knipknap:
You are speeed!
On first glance it looks ready to merge to me!

I will give it a deeper test and look tomorrow!

knipknap:
I found an issue that the power is not scaled correctly in some cases, which unfortunately led to a whole can of worms. The problem is not in your code - it is in mine; the root cause is that the power levels in Ops are not normalized.

I am working on fixing that in your branch, but this ripples through the whole app. I hope to finish the fix today though :smile:

heathkh:
Oh that's interesting... I just started another branch last night fixing an issue of lower power than expected with the engraved depth mode. The raster mode worked fine but the engraved depth mode was sending @ max S255 command when it should have been S1000 for my machine.

knipknap:
> I just started another branch last night fixing an issue of lower power than expected with the engraved depth mode. The raster mode worked fine but the engraved depth mode was sending @ max S255 command when it should have been S1000 for my machine.

That sounds exactly like the issue. It should be fixed with a3ec1ca.

I made some minor additional cleanups and worked in the direction of migrating the simulation overlay to use the CairoEncoder, but there are still some parts of that missing. In any case, I think this can be done later, so I am merging this branch!

Thanks a lot again for the excellent implementation, it is really nice to use!

For now I plan on merging #41, and then try to fix the abysmal (unusable) performance of rendering depth-engraved paths on full color images in the 3D canvas...

PR #44, opened 2025-10-04
normalize frame power to 0.0-1.0 range in frame job operation

heathkh:
Fix for this issue: https://github.com/barebaric/rayforge/issues/45

The frame job was failing because:
Power out of range: head.frame_power is stored as an integer (0-1000 for GRBL), but SetPowerCommand expects a normalized value (0.0-1.0). Fixed by dividing by head.max_power before passing to get_frame().
No active laser head: The encoder requires an active laser to be set before processing line commands. Fixed by creating a new Ops object with set_laser(head.uid) first, then appending the frame operations to it. This ensures the SetLaserCommand appears before the movement commands in the operation sequence.
The changes were made in [rayforge/machine/cmd.py:60-71]

TODO:
[  ] - Add a test that fails before this change and passes after.

### DIFF
--- a/rayforge/machine/cmd.py
+++ b/rayforge/machine/cmd.py
@@ -57,8 +57,10 @@ class MachineCmd:
                     self._editor.ops_generator,
                     context,
                 )
+                # Normalize frame_power (0-1000 scale) to 0.0-1.0 range
+                normalized_power = head.frame_power / head.max_power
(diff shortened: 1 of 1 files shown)


knipknap:
Damn. I hope you didn't burn a rectangle into your machine.

PR #49, opened 2025-10-11
Experiment - Moving mkdocs.yml into website dir.

heathkh:
Changes:
- Moves the mkdocs.yml file out of the root and into the website subdir to keep website related things together.
- Moves the mkdocs "content" files into a website/content subdir to seperate content from configuration.
- Not pixi expert, but inlined mkdocs cmd to remove the need for extra internal tasks (noob here, but like simpler)


knipknap:
Ha, that is almost exactly what I just did!

PR #57, opened 2025-10-14
Improve motion when engraving small vectors over serial

fllmn:
Noticed the engraver would start shaking when engraving small vectors. It turned out that the engraved is starved of commands if the time taken to process was lower than ~0.1.

Not a 100% sure about this but is seems to work fine when testing. Happy to provide logs if needed.

Why I think this works: Both the command processing and connection monitoring call `async with _lock` which should cause them to wait for the lock to become available. As soon as one has finished it will yield when trying to reacquire the lock
https://github.com/barebaric/rayforge/blob/6a2111ac48213f5fbcb8cc09d0a723e55c6c318b/rayforge/machine/driver/grbl_serial.py#L177-L179

https://github.com/barebaric/rayforge/blob/6a2111ac48213f5fbcb8cc09d0a723e55c6c318b/rayforge/machine/driver/grbl_serial.py#L229-L231

Alternatively having a lower sleep time will also help

### DIFF
--- a/rayforge/machine/driver/grbl_serial.py
+++ b/rayforge/machine/driver/grbl_serial.py
@@ -276,9 +276,6 @@ class GrblSerialDriver(Driver):
                             )
                             self._job_running = False
 
-                # Release lock briefly to allow status polling
-                await asyncio.sleep(0.1)
-
             except asyncio.CancelledError:
                 logger.info("Command queue processing cancelled.")
                 break


knipknap:
Thanks for debugging this! Does it also work stutter-free when you just change the wait time to 0?

                await asyncio.sleep(0)

I feel that may be a little safer. On GRBL, status updates cannot be made to work smoothly anyway, but it may be better than nothing that way.

fllmn:
> Does it also work stutter-free when you just change the wait time to 0?

Yea, I'm fairly sure it will. I I tried 0.001 which seem to remove almost all stuttering. I'm away for few days so I'll give it a try next week.

knipknap:
I now think there is a more fundamental change needed. The driver should not wait for "ok" at all; it should send as much as it can to keep the 127 byte Grbl buffer, filled, and track separately how full the buffer is, updating it whenever an "ok" arrives. This is a more substantial change, but the only way for it to work more reliably.

knipknap:
I just merged a new Grbl driver into main that supports the Grbl streaming protocol which should fix this issue. Tracking in #61. If this works for you, then I think this PR can be closed.

fllmn:
Great stuff! I'll have a go at it later today

PR #58, opened 2025-10-14
Update appdata for Flathub linting

fllmn:
Sorry for labouring this for too long. There's been a problem with screenshot and icon mirroring at flathub but resolved with a changes to the build command

From what I can test this fixes all linting errors but I'm not sure if this works with Snap.

Further on after this is merged I we should be able to run the pipeline and then commit the updated manifest


knipknap:
Looks like the snap built fine!
Sorry about the screenshot link, I broke that when moving it to the website...

PR #59, opened 2025-10-15
i18n: update french translations and template

yoyey:
Updated the messages.pot template to include new strings and refreshed the French translation (fr.po).
Other locales were left untouched to avoid accidental removals.

Some translations may need further review or adjustment depending on the intended meaning in context.


knipknap:
Merci!

PR #60, opened 2025-10-17
Feature: Add file drag-and-drop functionality to import images

heathkh:
Add drag-and-drop functionality to the Rayforge canvas, allowing users to import files by dragging them from their file manager and dropping them onto the WorkSurface. Imported items will be centered at the drop location on the canvas. This feature is already documented in the user documentation but not yet implemented.  Also adds support for pasting images from the system clipboard. 

**Key features:**
- Canvas-level drop target for precise positioning
- Coordinate transformation (widget pixels → world mm)
- Visual feedback overlay
- Batch import configuration
- System clipboard paste integration

### DIFF
--- a/rayforge/workbench/drag_drop_cmd.py
+++ b/rayforge/workbench/drag_drop_cmd.py
@@ -91,8 +91,8 @@ class DragDropCmd:
         Called when drag leaves the canvas.
         Uses delayed removal to handle spurious leave events from GTK4.
         """
-        # Don't immediately hide - GTK4 sometimes fires spurious leave events
-        # during active drags. Delay gives time for motion/drop events to occur.
+        # Don't immediately hide - GTK4 sometimes fires spurious leave
+        # events during active drags. Delay allows motion/drop to occur.
         logger.debug("Drag leave signal received, scheduling delayed hide")
         GLib.timeout_add(100, self._delayed_hide_overlay)
 


knipknap:
Nice addition, it looks awesome!

May I suggest some refinements on the code organization?

- Should we keep design docs in the source tree? I would propose to keep them only in the PRs, as design changes quickly and I would try to avoid to maintenance burden.
- In general I try to avoid adding code into the mainwindow.py and surface.py. As central objects they are "complexity magnets" and are already very large. The general pattern I try to use is to put the implementation of dedicated functions into `*_cmd.py` files - in the workbench module for UI code, and in the doceditor module for backend-only code. So probaby a drag_drop_cmd.py or something in the workbench makes sense?
- Any reason to limit the feature to specific image formats? Why not support any mime type that our importers support?
- _import_image_from_clipboard() Looks like it needs some decomposition, the nesting and complexity is quite high.

Yeah, so most of it is just shifting code around.

rayforge/mainwindow.py
```diff
@@ -1205,12 +1205,156 @@ def on_copy_requested(self, sender, items: List[DocItem]):
...
         """
         Handles the 'paste-requested' signal from the WorkSurface.
+        Checks for image data on system clipboard first, then falls back
+        to workpiece paste. After pasting an image, the clipboard is
+        cleared so subsequent pastes use workpiece data.
         """
+        # Priority 1: Check if system clipboard contains image data
+        clipboard = self.get_clipboard()
+        formats = clipboard.get_formats()
+        has_image = formats.contain_mime_type("image/png") or \
```
knipknap: Either accept any support format by checking the existing `image.importer_by_mime_type`, or define a `mime_formats_that_support_drag_drop` map in the image module, if there is a good reason to limit this

rayforge/workbench/surface.py
```diff
@@ -1179,3 +1186,247 @@ def set_simulation_mode(
...
+    def _find_parent_overlay(self):
+        """Find the Gtk.Overlay parent that contains this canvas."""
+        widget = self.get_parent()
+        while widget:
+            if isinstance(widget, Gtk.Overlay):
+                return widget
+            widget = widget.get_parent()
+        return None
+
+    def _on_file_dropped(
```
knipknap: This function is also quite large, could it be decomposed?

heathkh:
Thanks for the thorough review! I've addressed all of your feedback:

## 1. Design docs in source tree ✅

**Agreed and removed.** I've deleted `docs/design/drag-and-drop-import.md` from the source tree.
## 2. Code organization - avoid complexity in mainwindow.py and surface.py ✅

** Refactored.** I've created a new `rayforge/workbench/drag_drop_cmd.py` module following your established `*_cmd.py` pattern. This removed:
- **147 lines from mainwindow.py** (the entire `_import_image_from_clipboard()` method and local imports)
- **248 lines from surface.py** (all drag-drop methods: `_setup_file_drop_target()`, `_on_file_dropped()`, overlay management, etc.)

The new `DragDropCmd` class encapsulates all drag-and-drop and clipboard paste functionality. MainWindow now simply initializes it:

```python
self.drag_drop_cmd = DragDropCmd(self, self.surface)
self.drag_drop_cmd.setup_file_drop_target()
```

This keeps the complexity out of your central UI objects and follows the same pattern as `ViewModeCmd` and `SimulatorCmd`.

## 3. MIME type support ✅

**Great suggestion!** I've replaced the hardcoded MIME type lists with dynamic checks against `importer_by_mime_type` from the image module:

```python
from ..image import importer_by_mime_type

# Check if we support this MIME type
if mime_type not in importer_by_mime_type:
    logger.warning(f"Unsupported file type: {mime_type} for {file_path}")
    continue
```

Now the feature automatically supports **any MIME type that our importers support** (SVG, DXF, PDF, PNG, JPG, BMP, etc.) without needing updates when new importers are added.

## 4. Decompose `_import_image_from_clipboard()` ✅

**Decomposed into 6 focused methods** to reduce nesting and complexity:

1. **`_import_image_from_clipboard()`** - Entry point, starts async clipboard read
2. **`_save_and_import_texture(texture)`** - Background thread coordinator
3. **`_save_texture_to_temp_file(texture)`** - Isolates file I/O, returns Path or None
4. **`_import_temp_file_and_cleanup(temp_path)`** - Main thread import operation
5. **`_cleanup_temp_file(temp_path)`** - Resource cleanup helper
6. **`_show_clipboard_error()`** - Error UI helper

This reduced nesting from 4-5 levels down to 1-2 levels max. Each function has a single, clear responsibility and is much easier to test and maintain.

---

### Additional improvements made:

- **Converted to f-strings** in `file_cmd.py` and `import_handler.py` (gettext supports f-strings!)

Let me know if you'd like any other refinements!

knipknap:
Beautiful! I will give it a thorough test later - code wise, it looks ready to merge!

knipknap:
I tried and it seems to work if I let go of the image quickly. But if I hover for a second, then the mainwindow just disappears. The only logs:

```
2025-10-17 21:26:13,616 - rayforge.workbench.drag_drop_cmd - DEBUG - Drop overlay added to parent Gtk.Overlay
2025-10-17 21:26:13,654 - rayforge.workbench.drag_drop_cmd - DEBUG - Drop overlay removed
^C
```

heathkh:
I can reproduce.  Testing a fix.

heathkh:
Pushed a proposed fix... Please test to confirm.

# Root Cause
GTK4 fires spurious "leave" signals during active drag operations when you hold the drag for several seconds. The original code immediately removed the overlay widget when the leave signal fired, which caused crashes because:
The drag operation was still active in GTK4's internal state
Something in GTK4's drag system tried to access the overlay widget
The widget no longer existed → crash
The fix: Add a 100ms delay before hiding the overlay. This allows the drag system to properly complete (drop or actual leave) before removing the widget.

knipknap:
Thanks, this fixes the issue for me, too.

Now I did get one crash while dragging an SVG, then choosing "direct vector import", but am unable to reproduce it again. Will double check tomorrow if anything may be going on.

knipknap:
Merged, Thank you!

PR #62, opened 2025-10-20
Add file for Flathub verfication

fllmn:
For Flathub verfication, see [comment](https://github.com/flathub/flathub/pull/6990#issuecomment-3423305764)

Never used mkdocs so a wild guess from my side


knipknap:
I created this file directly in the website repo. This link now does not give 404:


https://rayforge.org/.well-known/org.flathub.VerifiedApps.txt

Is it correct that this has no content?

fllmn:
Thanks for adding. I'll close this.

Most of the examples I found were all empty but it looks like a verification id should be added once the application is published. https://docs.flathub.org/docs/for-app-authors/verification#im-publishing-an-app-on-flathub-how-do-i-get-it-verified

PR #74, opened 2025-11-23
Fixed a bug where .dxf files would not show up when selected for import

h0witzer:


knipknap:
Thanks - I just merged this in a commit that removed some AI-leftovers :smile:

Issue #70, opened 2025-11-17
macOS Support


pgilfernandez:
It would be nice to have macOS builds.

I guess it's possible, we just need somebody to deal with it right?

Anyway, this issue acts as a reminder or shout out for a brave developer that would like to try it 😄

Thanks!

knipknap:
Yes, it should be possible. The app is already fully cross platform, and no code changes should be required. I am guessing that the main hurdle is getting libadw to run, but I have no Mac to try.

pgilfernandez:
Hi!

I took some time to work with AIs (2 weeks) and get a working Rayforge build for macOS, it's not being easy but not it not only opens but I'm able to load a svg and work with it:

<img width="1456" height="952" alt="Image" src="https://github.com/user-attachments/assets/f41bfca7-d01a-4629-a90d-1ad2544182f6" />

I didn't try to burn it on a machine as I have no a GLBR available (just a Ruida controller one), so the only issue I easily identify is that there is no 3D preview of my work:

<img width="1456" height="952" alt="Image" src="https://github.com/user-attachments/assets/0112e2aa-f79e-42f1-ab3a-0304fa237175" />

I'm keep you updated with my progresses but, as said, it's mainly AI work, I'm the guide and prompter, I don't know if it is adding junk or unnecessary code... but ATM this is better than nothing 😉

If you want me to try something to test it let me know.
Cheers

knipknap:
Wow, that is cool! MacOS support was requested by many! How is this built? Can this be done in a Github pipeline?

I found out the hard way that AIs are really bad at writing shaders and 3D code in general, they get too confused with coordinate systems. But Amazing that GL runs already.

Another possible source of this not rendering could be the textures produced by the pipeline. For debugging, I recommend switching the "show travel paths" toggle on, then you can see of vertices (not textures) are drawn at least.

pgilfernandez:
I'm working locally with a macOS environment, I'm trying to follow your scheme with Linux and Windows... I guess it will be able to do it in a GitHub pipeline but it's soon to "port" it there.

Now I get the 3D view, but looks like it's mirrored:

<img width="1456" height="952" alt="Image" src="https://github.com/user-attachments/assets/d80702b0-6a33-4fa8-a0d4-9f8435992caf" />

[...]

PR #83, opened 2025-12-04
Improve scaling behavior

pgilfernandez:
While working on the [macOS support ](https://github.com/barebaric/rayforge/issues/70) I feel like the scaling behavior is not natural so this is what I did:

- Corner handles honor the “Fixed Ratio” preference by default; holding `Shift` inverts that preference (locked → freeform, freeform → locked).
- Edge (side) handles always do non-proportional scaling, regardless of `Shift` or the "Fixed Ratio" toggle.
- The "Fixed Ratio" toggle in the properties panel now shares its state with the 2D canvas so both UI and drag interactions stay consistent.

Try it and let me know if you like it 😄

### DIFF
--- a/rayforge/doceditor/editor.py
+++ b/rayforge/doceditor/editor.py
@@ -83,6 +83,9 @@ class DocEditor:
             self._on_processing_state_changed
         )
 
+        # Global preference for maintaining aspect ratio during transforms.
+        self.aspect_ratio_locked: bool = True
+
         # Instantiate and link command handlers, passing dependencies.
         self.asset = AssetCmd(self)
         self.edit = EditCmd(self)
--- a/rayforge/doceditor/ui/item_properties.py
+++ b/rayforge/doceditor/ui/item_properties.py
@@ -99,7 +99,7 @@ class DocItemPropertiesWidget(Expander):
 
         # Fixed Ratio Switch
         self.fixed_ratio_switch = Adw.SwitchRow(
-            title=_("Fixed Ratio"), active=True
+            title=_("Fixed Ratio"), active=self.editor.aspect_ratio_locked
         )
         self.fixed_ratio_switch.connect(
             "notify::active", self._on_fixed_ratio_toggled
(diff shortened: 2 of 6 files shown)


knipknap:
Sorry, this one flew under my radar, the notification got lost in other mails. Unfortunately the code base has moved so much that the PR no longer applies.

In general, I tried to replicate the behavior of the Inkscape canvas. To me it would seem unexpected if the corner handles force aspect ratio by default - Inkscape does not follow this semantic, enabling forced aspect only through modifier keys.

Otherwise I think resolving the conflict between the aspect toggle in the properties and the mouse behavior is a good idea.

pgilfernandez:
Hi!

Well, don't worry, here you have the idea in case you want to implement it.

I understand that you were trying to replicate Inkscape behavior but think that it is a design software where it's normal to transform, deform, etc. your vectors while Rayforge is focused on burning so it is supposed that vector will get imported in their final sizes and, if the case, users will probably want to proportionally scale but not deform the designs.... don't you think?

That's why I turned the default to proportionally scale 😉

RobinWhitfield:
Just adding that I do think that you'd expect resizing to maintain aspect ratio by default.

It's caught me out a couple of times already.

knipknap:
I think it depends on what other software you use. I would definitely get tripped up by locked aspect, I have never used any software that does - it appears the majority locks aspect only with a modifier key.

Though IMO a strong argument is that Lightburn locks aspect by default, where many users may be coming from...

image

RobinWhitfield:
> I would definitely get tripped up by locked aspect

When you say tripped up, what do you mean? Do you mean that you'd instinctively hold shift and have it not maintain the aspect ratio?
As there are side drag controls to me at least it feels mostly superfluous even letting a corner drag affect the aspect.

On a related note, I'd expect the "Fixed Ratio" toggle on the right to affect this, and not just the numerical values. Maybe "Maintain Aspect Ratio" would be a better wording too - perhaps with a "reset to original aspect ratio" adjacent to the toggle in case you accidentally adjust it.

If you think that some people would want it reversed (as it is right now), perhaps that could be a software preference that will default the above mentioned toggle to whatever the software preference is. But I think I'd still have it fixed by default.

knipknap:
Based on a Discord vote I am closing this issue:

image

PR #110, opened 2026-01-04
Added a little tip for svg when import failed

PastaLaPate:
very basic message


knipknap:
Great idea! I just added it a little differently, to ensure it shows up in the UI and not just on CLI. It looks like this:

image

knipknap:
The default offset is also fixed as proposed! With e0a85b994bf03f1389a0fbee81b0a7bf3ba0975d

PR #117, opened 2026-01-13
Update metainfo

fllmn:
Flathub has started [requesting](https://github.com/flathub/org.rayforge.rayforge/pull/8#issuecomment-3736095563) a `vcs-browser` tag to link to source repository. So far just a warning but could become an error in the future.

Also updated the homepage


knipknap:
Perfect, Thank you!

PR #131, opened 2026-02-02
Replace mimetype with provides/mediatype tags

fllmn:
Flathub fails validation, despite it being described as deprecated. :shrug: 

``Appstream: 'E:org.rayforge.rayforge.metainfo.xml:mimetypes-tag-deprecated:37 The toplevel `mimetypes` tag is deprecated. Please use `mediatype` tags in a `provides` block instead to indicate that your software provides a media handler for the given types.'``


knipknap:
Oh, awesome, I saw the failure yesterday night but did not have time to fix it yet. Thanks for the PR!

Issue #135, opened 2026-02-07
Minimal gui size is too large than screen


ocoquet-afk:
1.0.2: can't resize the gui to fit in the screen size, minimal size of gui is too large.

knipknap:
I am afraid this issue does not have enough information to go on. Any more detail / screenshot?

ocoquet-afk:
can't give you a screenshot...

Try to redim the gui window and you'll see what i mind, you can't resize the window width under 1446 pixels, my screen is 1250 pixels width :)

the height is ok, approximatively 100 pixels

regards
Olivier

fllmn:
It seem like the minimum width is defined by the amount of buttons, and the name of the machine. 

<img width="1858" height="102" alt="Image" src="https://github.com/user-attachments/assets/faefd3bb-451e-4ff5-899d-dddbff0e866d" />

I imagine the buttons could be grouped depending on window size and the machine name could also be limited, at least what's being presented in the GUI

knipknap:
Ooh, good point! I will check... currently deep in a week long pipeline debugging session, I can improve this when I have that figured out.

knipknap:
I just pushed this:

<img width="1395" height="109" alt="Image" src="https://github.com/user-attachments/assets/fbcb7b83-8bcc-4957-8a9d-e9561a70a6a6" />

Does this resolve the issue?

PR #136, opened 2026-02-08
Flexible toolbar layout

fllmn:
Discussed in #135, the toolbar currently sets the minimum width of the application. I'm not convinced this is the best way to solve but I wanted to give Gemini CLI a go (it didn't do that well), anyhow

This collapses the machine controls when the UI is narrowed down.

From 
image

To

image

The two issues I see are, machine controls are the most useful buttons (IMO) and making them less accessible is disservice to the user, secondly the "trigger" is based on a fixed width, ~1400px which is always a poor choice as it would have to be updated any time buttons change.

What could be interesting to discuss in case

1. What buttons should be grouped and hid when width is narrowed down
2. How can the toolbar manage its width to decide if it needs to collapse to fit the window.

### DIFF
--- a/rayforge/ui_gtk/mainwindow.py
+++ b/rayforge/ui_gtk/mainwindow.py
@@ -186,6 +186,9 @@ class MainWindow(Adw.ApplicationWindow):
         else:
             self.set_default_size(1100, 800)
 
+        # Callback on window width change
+        self.connect("notify::default-width", self._on_width_changed)
+
         # HeaderBar with left-aligned menu and centered title
         self.header_bar = Adw.HeaderBar()
         vbox.append(self.header_bar)
@@ -532,6 +535,9 @@ class MainWindow(Adw.ApplicationWindow):
 
         # Trigger startup tasks when window is shown
         self.connect("map", self._trigger_startup_tasks)
+
+    def _on_width_changed(self, widget, param):
+        self.toolbar._on_width_changed(widget, param)
 
     def _trigger_startup_tasks(self, widget):
         """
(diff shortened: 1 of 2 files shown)


knipknap:
> What buttons should be grouped and hid when width is narrowed down

Right now the buttons are ordered in a "process" order, i.e. left are the icons that are likely to be used when opening the app, then viewing, editing/aligning, then executing. So the order does not indicate importance. I think it should be possible to assign an "importance" to buttons, selectively hiding individual tool buttons.

Another option would be to collapse the "view" icon group into a stack like the one you showed, but only if there is not enough space available.

> How can the toolbar manage its width to decide if it needs to collapse to fit the window.

I suspect that the question "is enough space available?" is hard to answer, because the button size depends on the theme, icon, possibly font size, custom CSS, etc. and is not known ahead of time.
It can probably be done with "hacks" that detect overflow after drawing, but that would flicker... 

There is also [Adw.WrapBox](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.8/class.WrapBox.html), though a two-line toolbar may also not be ideal...

knipknap:
I realized that the problem is actually worse: In your screenshot, in between the machine selector and the WCS selector, that is where machine errors would currently be displayed. So when there is such an error the space limitation is much worse.

So maybe we need another approach entirely, moving the machine selector and errors somewhere else. The control panel would be a candidate, but I prefer to keep that hidden and that would not be good for error displays. The machine selector could also be moved into the title bar, but then where to put the error message... phew.

PR #141, opened 2026-02-18
macOS startup and shared-memory compatibility

pgilfernandez:
Well, a little step forward: after applying this two little commits I was able to open Rayforge: this PR includes two targeted fixes required to run `main` reliably on macOS.

### 1) `rayforge/ui_gtk/shared/keyboard.py`
On macOS, some GDK builds do not expose `Gdk.ModifierType.MOD2_MASK`.
The previous code accessed `MOD2_MASK` directly at import time, causing:

- `AttributeError: type object 'ModifierType' has no attribute 'MOD2_MASK'`

This prevented app startup on macOS in affected environments.

**Fix:** build the primary modifier mask dynamically using only available
modifier flags (`META_MASK`, `SUPER_MASK`, `MOD2_MASK` when present), with a
safe fallback to `CONTROL_MASK` if none are available.

### 2) `rayforge/pipeline/artifact/store.py`
Shared-memory names were built from long `creator_tag` values. On macOS,
`shm_open` enforces strict name-length limits, which caused runtime failures:

- `OSError: [Errno 63] File name too long`

This appeared during pipeline/view artifact creation and broke rendering flows.

**Fix:** generate short, sanitized SHM names (`rf_<short_tag>_<short_uuid>`),
keeping them unique while staying within macOS limits.

## Notes from macOS validation

- App startup now succeeds past the previous `MOD2_MASK` crash.
- `File name too long` SHM errors no longer reproduce.
- Remaining `vips-heif` warnings are environment/library ABI issues
  (`libx265`/`libheif`) and are unrelated to these code changes.

Could this PR be merged as it is or do you suggest any other approach?

What will be the next step? I would say that I could do a test and send you the log with the warnings and errors I find...

BTW, at a first sight, I have no visuals of either a SVG file or a sketch, I just get the final laser cuts (magenta paths) so I can't even move or edit the workpieces:

Screenshot 2026-02-18 at 22 42 56

### DIFF
--- a/rayforge/pipeline/artifact/store.py
+++ b/rayforge/pipeline/artifact/store.py
@@ -46,6 +46,7 @@ from __future__ import annotations
 import re
 import uuid
 import logging
+import zlib
 from typing import Dict, Optional, TYPE_CHECKING
 from multiprocessing import shared_memory
 import numpy as np
@@ -60,10 +61,49 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-def _build_shm_name(creator_tag: str) -> str:
+_CREATOR_TAG_ALIASES = {
+    "job": "job",
+    "step": "step",
+    "workpiece": "wp",
+    "workpiecechunk": "wpc",
+    "workpieceview": "wpv",
+    "viewlivebuffer": "vlb",
+    "render": "rnd",
+    "ops": "ops",
+}
+
+
+def _short_creator_tag(creator_tag: str) -> str:
     clean_tag = re.sub(r"[^a-z0-9]+", "", creator_tag.lower())
-    short_tag = clean_tag[:4] if clean_tag else "unk"
-    return f"rf_{short_tag}_{uuid.uuid4().hex[:16]}"
+    if not clean_tag:
+        return "unk"
+
+    direct = _CREATOR_TAG_ALIASES.get(clean_tag)
+    if direct:
+        return direct
+
+    if clean_tag.endswith("chunk"):
+        return "wpc"
+    if clean_tag.endswith("render"):
+        return "rnd"
+    if clean_tag.endswith("ops"):
+        return "ops"
+    if "workpieceview" in clean_tag:
+        return "wpv"
+    if "viewlivebuffer" in clean_tag:
+        return "vlb"
+    if "workpiece" in clean_tag:
+        return "wp"
+    if "step" in clean_tag:
+        return "step"
+    if "job" in clean_tag:
+        return "job"
+
+    return f"u{zlib.crc32(clean_tag.encode('ascii')):08x}"
+
+
+def _build_shm_name(creator_tag: str) -> str:
+    return f"rf_{_short_creator_tag(creator_tag)}_{uuid.uuid4().hex[:16]}"
 
 
 class ArtifactStore:


knipknap:
Looks great - though why rename the SHM instead of using a short name to begin with?

About this: "Remaining vips-heif warnings are environment/library ABI issues (libx265/libheif) and are unrelated to these code changes" - that probably explains the missing SVG rendering?

pgilfernandez:
> though why rename the SHM instead of using a short name to begin with?

AI answer this way:
```
We are not renaming an existing SHM object after creation (POSIX SHM can’t be safely “renamed” in place).
What changed is the **name generation strategy at creation time**.

Why this approach:

1. Minimal-risk fix: only `ArtifactStore` changed, no pipeline call-site changes.
2. Keeps traceability: we still keep a short `creator_tag` prefix (`rf_work_...`, `rf_view_...`) for debugging.
3. Guarantees macOS-safe length while preserving uniqueness (`uuid` suffix).
4. Cross-platform safe: same lifecycle semantics, just safer identifiers.

So effectively: it is already “using a short name to begin with,” centralized in one place.
```

> About this: "Remaining vips-heif warnings are environment/library ABI issues (libx265/libheif) and are unrelated to these code changes" - that probably explains the missing SVG rendering?

I'll have a look at try trying to fully use it and I will let you know

knipknap:
That is true, what I mean is: Instead of

    clean_tag = re.sub(r"[^a-z0-9]+", "", creator_tag.lower())
    short_tag = clean_tag[:4] if clean_tag else "unk"

We could just define the creator_tag short enough and avoid string cropping that may cause conflicts in the future.

pgilfernandez:
> We could just define the creator_tag short enough and avoid string cropping that may cause conflicts in the future.

OK, I understand and agree, tell me if you want me to include the new naming here or you prefer adding it as an normal commit. If here tell me what do you want 😊

knipknap:
I have no real preference, maybe "step", "wp" (for workpiece), "wpv" (workpiece view), "job"?

pgilfernandez:
In the meantime I was able to fix my libx265/libheif issues: I had to deactivate just HEIF... do Rayforge needs it?
Anyway, the non previewing the SVG issue was not that, if fixed in the SVG fallback:

On macOS environments where `libvips` is built without `svgload_buffer`,
Rayforge uses the Cairo/Rsvg fallback path.

The fallback previously called:

- `Rsvg.Handle.render_document(ctx, Rsvg.Rectangle())` with an empty/default rectangle.
In this runtime, that produces an empty render target (transparent/blank), so imported SVGs could be invisible on the 2D canvas even though vector operations/pipeline data were still generated.

Technical change, in `rayforge/image/svg/svg_fallback.py`:

1. Require the correct GI binding version explicitly:
   - `gi.require_version("Rsvg", "2.0")`
2. Render with an explicit viewport rectangle:
   - `x=0`, `y=0`, `width=target_width`, `height=target_height`
3. Keep the existing Cairo -> pyvips conversion flow unchanged.

knipknap:
Awesome, implementing it this way makes sense!

pgilfernandez:
> I have no real preference, maybe "step", "wp" (for workpiece), "wpv" (workpiece view), "job"?

OK then, I updated the SHM naming strategy to use explicit short creator tags (e.g. `job`, `step`, `wp`, `wpc`, `wpv`, `vlb`) instead of truncating arbitrary strings with `[:4]`.

This keeps names macOS-safe while reducing collision risk from truncation. For unknown tags, we now use a stable hash-based token rather than cropping. The uniqueness guarantee still comes from the UUID suffix.

Is it fine like it is now? ... I think we have a conflict...

pgilfernandez:
Apart from that conflict, I'm glad to say Rayforge compiles back again in macOS, I'm able to create the bundle and pack it into a DMG for distribution.

The additional macOS only files I would need to add (here or in another PR?) are:

- `.gitignore` Ignore local/generated artifacts (`.mac_env`, `rayforge/version.txt`) so build state is not committed.
- `Rayforge.spec` Adds the macOS PyInstaller spec used to produce the `.app` bundle consistently.
- `hooks/hook-gi.repository.Gtk.py` Ensures PyInstaller uses the GTK 4 hook (instead of GTK 3 defaults) so GI resources are bundled correctly.
- `rayforge/resources/icons/rayforge.icns` Provides the native macOS app icon used by the `.app` bundle.
- `scripts/mac_setup.sh` Adds a macOS setup flow for environment/dependency preparation and local run instructions.
- `scripts/mac_build.sh` Adds the macOS build/bundle/dmg pipeline and bundle-time runtime dependency fixes.
- `scripts/mac_create_icon.sh` Automates `.icns` generation from the project SVG icon for reproducible macOS packaging.

Let me know how do you want me to proceed 😉

BTW, after this I will comment you some issues I found (some of them were fixed by AI in my branch but it makes sense to do it correctly here so I prefer comment them to you to define the best way to fix them) 😊

knipknap:
Cool, I will clean up some things and merge it soon! Thanks for figuring all this out!

Yes, the build stuff should definitely be a separate PR. And sure, if you found other issues separate PRs are always better ;-)

knipknap:
This is now in main with a62068950ae4222b99a668a8a29a082346b8ce80.

pgilfernandez:
> This is now in main with [a620689](https://github.com/barebaric/rayforge/commit/a62068950ae4222b99a668a8a29a082346b8ce80).

@knipknap, and what about the changes in `keyboard.py` and `svg_fallback.py`?

pgilfernandez:
> > This is now in main with [a620689](https://github.com/barebaric/rayforge/commit/a62068950ae4222b99a668a8a29a082346b8ce80).
> 
> @knipknap, and what about the changes in `keyboard.py` and `svg_fallback.py`?

Forget this comment, I saw them in different commits, I guess you cherry picked them 😉

PR #143, opened 2026-02-20
fix(camera): adjust offset bounds and improve device scanning

MausRundung:
- Change X/Y offset spin rows from 0-10000 to -100-100 with 0.01 step increment for finer control
- Replace subprocess camera scanning with direct OpenCV backend iteration to avoid crashes
- Remove Windows-specific OpenCV environment variable overrides
- Reduce log level for device scanning from info to debug to reduce noise

### DIFF
--- a/rayforge/camera/controller.py
+++ b/rayforge/camera/controller.py
@@ -1,4 +1,3 @@
-import os
 import sys
 import time
 import threading


MausRundung:
The automated checks are failing, but my changes were minimal and didn't touch the imports. I’m currently checking the logs to see if this is a environment issue.

knipknap:
Thanks for the PR! The camera part looks good, but I do not understand the offset part. The offset is supposed to mean "the real usable part of my machine is a subset of the machine dimensions, and that subset is at x/y". So negative values would be logically impossible, because it would move the work area outside the machines physical dimension.

Also, why clamp it to 100? `lower=-100, upper=100`?

MausRundung:
The laser does not necessarily need to start at 0x, 0y. Mine, for example, has a resting and cleaning station at -25x and -12.40y because not my entire movable area is my working area...

Yes, technically I could just extend the working area, but then you would have unusable space in the workspace every time, and this way the G-code looks cleaner and is easier to understand.

MausRundung:
And why go to 100 instead of 10000? I don't think there is any reason to have an offset go that high, not even on industrial machines.

knipknap:
Depending on the machine's native units, the value entered in the offset fields represents micrometers, inches, meters - anything really (see issues like #4). So I think we could remove the clamps entirely, or at least leave them at a very high value.

Honestly I am split about the negative offset: Currently, it would not cause any issues, so this is an easy way to achieve what you want. The problem is the long term issues it may cause later on: The surface is currently designed to represent the "digital twin", i.e. a representation of the physical machine, and using work coordinate systems (e.g. G55 etc.) are the "correct" way to handle "sub-coordinate systems". That is why a while back I proposed to remove the offset setting entirely, as it is redundant to WCS. But I kept it for backward compatibility as some users thought it is easier for them to understand.

However, in the future we may want to add better machine visualizations to represent the physical machine, e.g. in a 3D model or something. The "usable subsurface" would then be drawn differently than the physical machine in the app. At that point, negative offsets will break the UI for anyone using them, and anyone using offsets would be forced to migrate to using WCS anyway.

So should we really add a concept that we know will break in the future?

MausRundung:
Instead of G-code, we could move the offset to the camera settings and add workspace deadzones. For now, though, the current method works well and likely meets the needs of most users.

knipknap:
I am confused. Neither WCS not the offset settings have any relation to the camera settings. Could you elaborate?

MausRundung:
The camera view is bound to the workspace. If the gantry's physical zero doesn't match the workspace zero, the camera overlay and the laser's actual strike point will be misaligned. Allowing negative offsets is a simple way to sync the physical hardware with the software's coordinate system without having to redefine the entire machine's travel area.

knipknap:
Ok, my understanding:

> The camera view is bound to the workspace.

The camera view is bound to the physical dimensions of the machine, as defined in the machine settings under hardware. The camera image can be aligned in this space by using the camera alignment dots.

> If the gantry's physical zero doesn't match the workspace zero

Here is where I stumble: "physical zero" I would define as the physical maximum extents of the machine, i.e. what is configured under "Hardware". "Workspace" would then be a subsection of that surface, i.e. if your build plate is smaller than what your axis can travel, then the workspace would define the bounds of that. In other words, the workspace is the "usable area", not the physical extents.

knipknap:
Thinking about this more I think it's best if I restructure the "Hardware" page and even add a diagram to explain what each configuration setting means. The general idea:

1. Replace "width" and "height" in the "Dimensions" section by maximum extents in the Axes section, explaining that they refer to the maximum extents of the axes regardless of the size of the actually workable area.
2. Remove X offset and Y offset. This will be replaced by what follows.
3. Add a new section "Work Surface". This allows to specify the bounds of the work surface, i.e. the workable area in world space defined by the axis extents.
4. Add a diagram that dynamically updates to visualize the result.
5. (optional, maybe longer term: update the main canvas visualization to make the distinction between axis extents and work surface clearer)
6. (option, longer term: add the ability to configure soft limits)

Since this is a more substantial change I propose to make an issue and I can work on it.
I would for now only merge the Camera part of your PR.

Thoughts?

MausRundung:
sounds good

knipknap:
The camera part is now in main. Thanks for the PR!

PR #148, opened 2026-02-24
macOS: fallback shader program compile without validation on macOS

pgilfernandez:
### Summary
This PR improves 3D canvas initialization robustness on macOS by handling early shader validation failures.

### Problem
On some macOS/OpenGL contexts, shader program creation fails during validation with:
- `ShaderValidationError: Current draw framebuffer is invalid`

This can happen when validation runs before a valid draw framebuffer is available, causing 3D initialization to fail.

### Change
In `rayforge/ui_gtk/canvas3d/gl_utils.py`:
- Keep the normal shader compile/link path.
- If `ShaderValidationError` occurs, retry `compileProgram(..., validate=False)`.

### Impact
- Prevents 3D initialization failure in affected macOS contexts.
- Preserves normal behavior when validation succeeds.
- Scoped to shader initialization; no changes to 2D pipeline or artifact generation.


knipknap:
Test passed - Thank you for the fix!

PR #149, opened 2026-02-24
macOS: shorten shared-memory creator tags for step artifacts on macOS

pgilfernandez:
### Summary
This PR fixes a macOS runtime failure when creating shared-memory artifacts for step assembly (Intel, Apple Silicon, and universal builds).

### Problem
Step artifact creation used creator tags that produced shared-memory names like `rf_step_render_*`, which can exceed the POSIX shared-memory name limit on macOS. This caused:
- `OSError: [Errno 63] File name too long`
- step assembly failures
- downstream preview/simulation errors (`Job dependencies are not ready`)

### Change
In `rayforge/pipeline/stage/step_runner.py`, creator tags were shortened:
- `step_render` -> `step_r`
- `step_ops` -> `step_o`

### Impact
- Prevents shared-memory name overflow on macOS.
- Stabilizes step assembly and preview/simulation flow.
- Runtime-only fix; no functional change to toolpath/ops logic.

### DIFF
--- a/rayforge/pipeline/stage/step_runner.py
+++ b/rayforge/pipeline/stage/step_runner.py
@@ -101,7 +101,7 @@ def make_step_artifact_in_subprocess(
 
     # 1. Store Render Artifact
     render_handle = artifact_store.put(
-        render_artifact, creator_tag=f"{creator_tag}_render"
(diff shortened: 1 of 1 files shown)


knipknap:
Perfect - thanks again for the fix!

PR #150, opened 2026-02-24
macOS CI

pgilfernandez:
There we go:

Now we have Intel, ARM and universal (combines both Intel and ARM) apps compiling without errors, there is an extra step that packages the universal one into a DMG... we could fine tune the artifacts we want to end up in the near future, at this point I prefer not to touch anything else...

The apps bundled with CI are not working 100%, they have different issues that I'm trying to fix (PR) in other PRs. The point here is to set a milestone and continue working out further problems even if we finally tag them as CI issues.

BTW, this is the real premiere of Rayforge under macOS 😄💪🏻

### DIFF
--- a/website/src/components/InstallGuide.js
+++ b/website/src/components/InstallGuide.js
@@ -78,6 +78,12 @@ const windowsMethods = [
   { id: 'developer', label: translate({ id: 'install.msys2.developers', message: 'MSYS2 (Developers)' }) },
 ];
 
+const macosMethods = [
+  { id: 'universal', label: 'Universal (Recommended)' },
+  { id: 'arm', label: 'Apple Silicon (M1/M2/M3)' },
+  { id: 'intel', label: 'Intel Mac' },
+];
+
 function OsSelector({ selectedOs, onSelectOs }) {
   return (
     <div className="install-os-selector">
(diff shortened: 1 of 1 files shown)


knipknap:
This is amazing, fantastic job! Thank you!

One suggestion is to extend the workflow by a "smoke test" - maybe not the whole test suite is needed, but at least test if the package can be installed and that `rayforge --version` works.

knipknap:
Another small request: Could you also update the installation instruction in `website/src/components/InstallGuide.js` for Mac?

pgilfernandez:
> This is amazing, fantastic job! Thank you!
> 
> One suggestion is to extend the workflow by a "smoke test" - maybe not the whole test suite is needed, but at least test if the package can be installed and that `rayforge --version` works.

done!

pgilfernandez:
> Another small request: Could you also update the installation instruction in `website/src/components/InstallGuide.js` for Mac?

Indeed, may I force push it here or a new commit?

knipknap:
> Indeed, may I force push it here or a new commit

Either one is fine with me :-)

pgilfernandez:
> > Indeed, may I force push it here or a new commit
> 
> Either one is fine with me :-)

OK, I had rebased the PR as you edited `InstallGuide.js`, so I added the new macOS builds and little explanation... It's the first time I use this way of website deploying so I may change/improve things when I see them online, or feel free to improve or change anything you want 😊

knipknap:
Thanks again!

PR #151, opened 2026-02-26
Fix G-code simulation preview line thickness scaling with zoom

pgilfernandez:
### Summary
This PR fixes a rendering issue in the 2D G-code simulation preview where
stroke thickness changed with zoom level (especially noticeable on macOS).

### Changes
- Updated `SimulationOverlay` to use constant device-pixel stroke width for:
  - travel/cut path rendering
  - scanline segments
  - laser head crosshair/circle indicator
- Added a robust fallback for environments without `cairo.Context.set_hairline`
  using `device_to_user_distance`.
- Added a UI test to verify hairline stroke mode is used when available.

### Why
Previously, stroke widths were defined in world units (mm), so zooming in/out
made lines appear too thick or too thin. This made path readability and the
laser position marker inconsistent.

### Validation
- Syntax-checked modified files with `python -m py_compile`.
- Added regression coverage in:
  - `tests/ui_gtk/elements/test_simulation_overlay.py`

### Notes
- No behavior changes to timeline/state logic.
- This change is focused only on visual stroke consistency in simulation overlay.

### DIFF
--- a/rayforge/ui_gtk/canvas2d/elements/simulation_overlay.py
+++ b/rayforge/ui_gtk/canvas2d/elements/simulation_overlay.py
@@ -312,13 +312,7 @@ class SimulationOverlay(CanvasElement):
         """
         Keeps stroke width at roughly one device pixel regardless of zoom.
         """
-        if hasattr(ctx, "set_hairline"):
-            ctx.set_hairline(True)
-            return
-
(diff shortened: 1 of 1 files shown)


pgilfernandez:
It looks like the Intel job got stuck randomly... could you relaunch it again please?

pgilfernandez:
Thanks!
It finally compiled it all and now it works as expected, thanks.

PR #164, opened 2026-03-09
Wire macOS app menu About, Settings, and Quit actions

pgilfernandez:
On macOS, the global application menu still exposed the standard `About`,
`Preferences`, and `Quit` entries, but they were not wired to the same actions
used by Rayforge's in-window menu:

Screenshot 2026-03-09 at 16 35 35

This change adds application-level actions
for `app.about`, `app.preferences`, and `app.quit` in
[rayforge/app.py](/Users/pablo/GitHub/rayforge/rayforge/app.py) and forwards
them to the active main window so the existing UI flows are reused.

As a result, `Rayforge > About Rayforge` now opens the same dialog as
`Help > About`, `Rayforge > Quit` exits through the same path as the main
window, and `Rayforge > Preferences` is handled as `Settings` on macOS and
opens the same window as `Edit > Settings`. The related keyboard accelerators
were also moved to the `app.*` action namespace, which is the correct scope for
the native macOS application menu.

### DIFF
--- a/rayforge/app.py
+++ b/rayforge/app.py
@@ -164,10 +164,14 @@ def main():
                 self.add_action(action)
 
         def _get_main_window(self):
+            from rayforge.ui_gtk.mainwindow import MainWindow
+
             window = self.get_active_window()
-            if window is not None:
+            if isinstance(window, MainWindow):
                 return window
-            return self.win
+            if isinstance(self.win, MainWindow):
+                return self.win
+            return None
 
         def _on_app_about(self, action, param):
             window = self._get_main_window()


pgilfernandez:
Hi @knipknap, could you stop and relaunch the intel CI? it got frozen... thanks!

knipknap:
Done

pgilfernandez:
Great, thanks!
It works as expected so it's now mergeable 😉

Screenshot 2026-03-10 at 10 39 18

knipknap:
It looks like the linter still fails:

```
/home/runner/work/rayforge/rayforge/rayforge/app.py
  /home/runner/work/rayforge/rayforge/rayforge/app.py:176:20 - error: Cannot access attribute "show_about_dialog" for class "Window"
    Attribute "show_about_dialog" is unknown (reportAttributeAccessIssue)
  /home/runner/work/rayforge/rayforge/rayforge/app.py:182:20 - error: Cannot access attribute "show_settings" for class "Window"
    Attribute "show_settings" is unknown (reportAttributeAccessIssue)
  /home/runner/work/rayforge/rayforge/rayforge/app.py:189:20 - error: Cannot access attribute "on_quit_action" for class "Window"
    Attribute "on_quit_action" is unknown (reportAttributeAccessIssue)
```

Probably "if window is None" needs to be replaced by

```
        from rayforge.ui_gtk.mainwindow import MainWindow
        if not isinstance(window, MainWindow)
```

PR #203, opened 2026-04-14
Add rotary axis support with TRUE_4TH_AXIS and AXIS_REPLACEMENT modes

### LINKED ISSUE #195: Ability to map Rotary Axis to Y
I am using a laser engraver setup that uses an external rotary module. My engravers ( a MecPow X3 Pro and a Longer Ray5) do not have a dedicates A-axis port. Instead, they require the user to disconnect the Y-axis motor and plug the rotary module directly into the Y-axis port.

But currently, the software outputs G-code using the A axis for rotary movements (or other between B, C, U, Z but no Y)

Image

**Proposed Solution:**
It would be a great improvement to add an option in the "Rotary Module" settings to select the output axis Y.
Furthermore, in the case of the Y-axis, which is a linear motion axis, it might be worth introducing a "Steps per mm/degree" to correctly scale the movement. If the rotary axis is connected to the Y-axis, the software often has to be tricked into considering the distance traveled (circumference of the object) compared to the standard linear mm.

knipknap:
## Summary

- Refactors rotary axis handling by introducing `AxisSet` and `AxisMapper` to generically map flat coordinates to a rotary axis in the pipeline
- Adds support for `TRUE_4TH_AXIS` and `AXIS_REPLACEMENT` (renamed from `PASSTHROUGH`) rotary modes, allowing switching X or Y for rotary (#195)
- 2D canvas now allows defining the rotary projection direction; 3D canvas and pipeline generically map flat-to-rotary
- Machine settings now offer settings for roller-type rotary axis
- Rebased on main with conflict in `canvas3d.py` resolved


knipknap:
This PR has now so many fixes compared to main that I consider it past the point of no return. Will merge and if there are still new issues (currently not aware of any) they will have to be resolved in main.

PR #211, opened 2026-04-18
add GRBL (Telnet) driver for networked grblHAL and ESP3D controllers

gyordanov:
Subclasses GrblSerialDriver and swaps SerialTransport for TelnetTransport.
Covers networked grblHAL and ESP3D controllers with telnet enabled — the
existing GRBL (Network) driver only targets ESP3D+WebUI3.

Relaxes GrblSerialTransport to accept any Transport (it only used the
base-class API anyway); .port returns "" for non-serial backends.

- New `GrblTelnetDriver` for networked grblHAL / ESP3D via raw telnet (port 23).
- Subclass of `GrblSerialDriver` — reuses the full protocol stack, only swaps the byte transport.
- `GrblSerialTransport` constructor now typed as `Transport` instead of `SerialTransport` (strict annotation relax, no behaviour change).

## Why a new driver
grblHAL exposes telnet, WebSocket, HTTP, and WebUI as independent services. The existing `GrblNetworkDriver` hard-requires the WebUI3 HTTP endpoints
(`[ESP420]`/`[ESP800]`/`[ESP400]`, `/command?commandText=...`), so it can't talk to a plain networked grblHAL or an ESP3D bridge.

## Test plan
- [x] lint + full test suite (2 preexisting unrelated failures)
- [x] smoke test against networked grblHAL
- [x] smoke test against ESP3D

### DIFF
--- a/rayforge/machine/driver/__init__.py
+++ b/rayforge/machine/driver/__init__.py
@@ -4,6 +4,7 @@ from .driver import Driver
 from .dummy import NoDeviceDriver
 from .grbl import GrblNetworkDriver
 from .grbl_serial import GrblSerialDriver
+from .grbl_telnet import GrblTelnetDriver
 from .smoothie import SmoothieDriver
 
 
@@ -34,5 +35,6 @@ __all__ = [
     "NoDeviceDriver",
     "GrblNetworkDriver",
     "GrblSerialDriver",
+    "GrblTelnetDriver",
     "SmoothieDriver",
 ]
--- a/rayforge/machine/driver/grbl_telnet.py
+++ b/rayforge/machine/driver/grbl_telnet.py
@@ -1,0 +1,106 @@
+import logging
+from typing import Any, Optional, cast
+from gettext import gettext as _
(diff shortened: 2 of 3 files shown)


knipknap:
Well this is quite cool. I am happy to see the implementation is so straight forward, it means the transport protocol abstraction is working well.

I am not sure what you mean by "lint + full test suite (2 preexisting unrelated failures)". I am not aware that there are any pre-existing failures.

gyordanov:
i think its something to do with my arch/setup ... i'll rerun and see what it prints but the lint/test run had identical results with and without the changes ...  pixi is fighting me right now - the only way to run this is to build an arch package - pixi does not not seem to like arch

gyordanov:
no failures -  all tests pass, it was git not fetching the tags, and then that freaking out addon check and making tests fail.

knipknap:
Great - Since all tests passed I merged it :-). Thank you!

gyordanov:
amazing ! thank you!

PR #224, opened 2026-04-22
macOS: Avoid importing GTK/UI modules from code paths that are also loaded by background workers

pgilfernandez:
On macOS, worker processes that import GTK/UI modules appear as extra Rayforge app instances in the Dock (bouncing like crazy and they never close, every new calculation creates new instances...)

So, I made these changes:

- Move `laser_essentials.worker` UI-only imports into command/action registration hooks.
- Move laser addon icon registration into the frontend entry point.
- Make `rayforge.ui_gtk` convenience exports lazy to avoid loading the 2D canvas when workers import 3D scene modules.

This way Rayforge works again fine in macOS 😄

### DIFF
--- a/rayforge/builtin_addons/rayforge-addon-laser/laser_essentials/frontend.py
+++ b/rayforge/builtin_addons/rayforge-addon-laser/laser_essentials/frontend.py
@@ -8,6 +8,7 @@ import gettext
 from pathlib import Path
 
 from rayforge.core.hooks import hookimpl
+from rayforge.ui_gtk.icons import register_icon_path
 from .widgets import PRODUCER_WIDGETS
 
 _localedir = Path(__file__).parent.parent / "locale"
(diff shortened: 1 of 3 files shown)


knipknap:
Thanks! While the idea of the fix looks good I think we can simplify by moving the hooks out of the worker completely. I just pushed this different change to main, can you try if that works?

pgilfernandez:
I just tested it and it work for calculations in the "2D view" but processes in the 3D view open instances... follow my PR to patch it too, thanks

knipknap:
Thanks - I am trying to avoid the getattr based workaround - it is an indication that something more fundamental is wrong. The problem is that the subprocess that prepares the 3D scene imports UI related modules from ui_gtk, i.e. UI and backend were not cleanly separated.

I now tried to solve this by moving the scene compiler from the UI to a backend module, as it does not actually use any UI code. That **should** make it safe for the compiler subprocess to import.

pgilfernandez:
It now works fine 😄

knipknap:
Awesome, Thank you reporting and testing!

PR #249, opened 2026-05-18
feat(device): add Sculpfun C1 engraver

BoboTiG:
Hello,

A patch to add the Scupfun C1 engraver to the list of known devices (this is this one: https://www.sculpfun.com/collections/laser-engraver/products/sculpfun-c1-mini-laser-engraver-3w).

I duplicated the `sculpfun-s30` file structure, and filed with what the software found when I added the machine manually (by copying only the relevant data saved in `~/.var/app/org.rayforge.rayforge/config/rayforge/dialects` + `~/.var/app/org.rayforge.rayforge/config/rayforge/machines`).

### DIFF
--- a/rayforge/resources/devices/sculpfun-c1/device.yaml
+++ b/rayforge/resources/devices/sculpfun-c1/device.yaml
@@ -13,8 +13,8 @@ machine:
   max_cut_speed: 10000
   heads:
   - max_power: 1000
-    frame_power_percent: 0.0
-    focus_power_percent: 0.0
+    frame_power_percent: 0.25
+    focus_power_percent: 0.25
     spot_size_mm:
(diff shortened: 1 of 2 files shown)


knipknap:
Awesome, Thanks for the PR! Perhaps setting a frame and focus power makes sense?

BoboTiG:
That's good on my side :)

BoboTiG:
The only nit-pick is the max cut speed. Being a 3w laser, even if the theoric value is 10,000; in reality it would be 400 or something like that. I guess this is OK to keep the theoric value.

I never used it to cut, so maybe someone later could improve the value, if needed.

knipknap:
Sounds good. Thanks again!

PR #259, opened 2026-05-25
Fix manual laser control routing

heiserhorn:
Description
Fixes the manual laser control widget so laser on/off commands are sent to the machine currently assigned to the widget, instead of relying on the globally active config machine.

Previously, clicking the manual laser button could enqueue set_focus_power() against the wrong machine context, which meant the selected laser did not turn on from the widget.

Changes

  - Allow MachineCmd.set_power() and MachineCmd.set_focus_power() to accept an explicit machine.
  - Update LaserControlWidget to pass its assigned machine when turning the laser on or off.
  - Add focused tests for explicit machine routing in MachineCmd.

Verification

  - python3 -m compileall passed for changed files.
  - git diff --check passed.
  - Could not run pixi run lint or targeted pytest locally because pixi and pytest are not available in this environment.

### DIFF
--- a/rayforge/machine/cmd.py
+++ b/rayforge/machine/cmd.py
@@ -369,7 +369,12 @@ class MachineCmd:
             key=f"macro-{macro_uid}",
         )
 
-    def set_power(self, head: "Laser", percent: float):
+    def set_power(
+        self,
+        head: "Laser",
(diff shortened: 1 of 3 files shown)


knipknap:
Thanks for the fix!

PR #268, opened 2026-06-01
fix documentation link

thehamstermuffin:
Noticed that the developer documentation was broken and I found the correct one on the website


knipknap:
Thank you!

PR #270, opened 2026-06-02
Add raygeo as submodule

thehamstermuffin:
Having Raygeo configured as a submodules prevents having dependency errors.

Noticed it because I couldn't run `pixi install` when following the developer documentation, with this change you only need to pull the submodule with  `git submodule update --init --recursive` after cloning the repository

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "external/raygeo"]
+	path = external/raygeo
+	url = https://github.com/barebaric/raygeo


thehamstermuffin:
I'm not sure how to update the documentation on the website

knipknap:
The reason I didn't add it as a submodule is that this would mean maintaining version references in two places: requirements.txt and pixi.toml already specify a version to use; since submodules are also pinned to a specific version they would need to be kept in sync. As far as I know, submodule references cannot be set to "latest", they must remain bound to a specific commit.

I am not sure how to manage that better.

thehamstermuffin:
I see, it seems your implementation is good. But is there any way to download the module automatically? I couldn't get `pixi install` to run until I manually downloaded raygeo. 

      Error:   × failed to solve the pypi requirements of environment 'default' for platform 'linux-64'
      ├─▶ error while canonicalization /home/demo/Documentos/GitHub/rayforge/external/raygeo
      ╰─▶ No such file or directory (os error 2)
      
I kept getting this message.

knipknap:
You are right, this is also not nice. I have not found a way that checks all the boxes yet...

PR #285, opened 2026-06-10
Add macOS Tahoe (Liquid Glass-style) icons

LordImmaculate:
Hello

I added the new kind of macOS app icons (.icon instead of .icns) introduced with macOS Tahoe. This will make sure the icon looks great in clear or dark icon modes.

If the user is not on macOS 26+, it will fall back to the original .icns icon.

I also added .DS_Store to the .gitignore, this is a file that gets created for each folder automatically by Finder, which should not be committed.


knipknap:
Thanks for the PR! Looks like it is currently still failing:

```
  File "/Users/runner/work/rayforge/rayforge/.venv/lib/python3.11/site-packages/PyInstaller/building/icon.py", line 34, in normalize_icon_type
    raise FileNotFoundError(f"Icon input file {icon_path} not found")
FileNotFoundError: Icon input file /Users/runner/work/rayforge/rayforge/rayforge.icns not found
```

LordImmaculate:
I think the reason this happened is because the liquid glass icons were added in macOS 26 (so the command to add these too), but the GitHub actions were running on macOS 14 and 15. It worked on my computer running macOS 27, so I changed it in the actions.

LordImmaculate:
Updating the runners to macOS 26 fixed it.

knipknap:
Awesome, Thanks!

PR #297, opened 2026-07-05
Add support for Longer Ray5 over Wifi

uwewoessner:
Longer unfortunately uses a slight variant of the GRBL network protocol
You can now configure this in the Machine/Network settings


knipknap:
Great, Thanks! The failing test should be fixed if you rebase on main. This is caused by Raygeo 0.14.0 mismatch; I just pushed a port to that.

PR #305, opened 2026-07-11
feat: add air assist toggle to material test grid

### LINKED ISSUE #304: Air Assist for material test grid
Would be nice to add the ability to enable air assist when producing a material test grid.   The only way I have found to do it is to manually force the pump on before generating the test grid, sending the M8 command to the machine before starting, and an M9 once done.  A simple on/off button in the settings would be nice.

And along similar lines, is there a reason why air assist is not on by default anyway?   It wouldn't hurt IMO if it's always enabled (and you have to disable it if needed) for most operations.

knipknap:
## Summary

Adds an air assist toggle to the material test grid settings. Previously the `MaterialTestCapability` had an empty varset, so there was no UI control for enabling air assist during material test jobs. Users had to manually send M8/M9 commands to the machine before and after generating a test grid.

The fix adds an `air_assist` `BoolVar` to the `MaterialTestCapability` varset, following the same pattern used by Cut, Engrave, and Score capabilities. The pipeline already handles air assist state via `_create_initial_ops()` when `settings['air_assist']` is present, so no changes to the producer or G-code generation are needed — the M8/M9 commands are automatically emitted by the encoder based on the air assist state.

Closes #304

### DIFF
--- a/rayforge/core/capability.py
+++ b/rayforge/core/capability.py
@@ -325,7 +325,15 @@ class MaterialTestCapability(Capability):
 
     @property
     def varset(self) -> VarSet:
-        return VarSet(vars=[])
+        return VarSet(
+            vars=[
+                BoolVar(
+                    key="air_assist",
+                    label=_("Air Assist"),
+                    default=False,
+                ),
+            ]
+        )
 
 
 class PWMCapability(Capability):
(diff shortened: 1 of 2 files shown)


knipknap:
This is now in main

PR #314, opened 2026-07-18
fix: settings-widget sensitivity/default-value bugs in step popups

vyvcodd:
### Summary

- The Line Spacing, Sample Interval, Line Interval (Material Test Grid), and
  Step Over sliders displayed a hardcoded `0.1mm` whenever the underlying
  step property was unset, instead of the laser-spot-size-derived value
  actually used at generation time — so the slider never reflected what was
  really going to be generated until the user manually touched it.
- Overscan's "Automatic Distance" switch was permanently greyed out:
  `_update_sensitivity()` computed the row's own sensitivity from its own
  active state (`enabled and not auto and not native`), and `auto` defaults
  to `True`, so the switch disabled itself from the moment it existed.
- Auto-calculated Overscan/Lead-In-Out distance silently shrank when
  disabling Auto and re-enabling it: `EngraveStep`/`ContourStep`/
  `ShrinkWrapStep`/`FrameStep` all seeded the initial distance from
  `machine.max_cut_speed`, but the widgets' own recalculation on re-enable
  uses `step.cut_speed` — typically far lower, and distance scales with
  speed².

### Changes

- `rayforge/ui_gtk/doceditor/step_settings/base.py`: add
  `get_selected_laser()` helper for widgets to compute real fallback values.
- `raster_widget.py`, `material_test_grid_widget.py`, `wavefront_widget.py`:
  use the actual selected laser's spot size as the displayed default
  instead of a hardcoded `0.1`.
- `overscan_widget.py`: `auto_row`'s sensitivity no longer depends on its
  own active state.
- `raster_step.py`, `contour_step.py`, `shrinkwrap_step.py`, `frame_step.py`:
  compute the initial auto distance from `step.cut_speed` (after the
  per-laser-capability defaults loop, which can itself set it), matching
  what the widgets recompute on toggle.

### Checks

- `pixi run test`: 5286 passed, 300 deselected
- `pixi run lint`: clean (2 pre-existing, unrelated pyright errors on
  `sys._MEIPASS` in `app.py`/`worker_init.py`)
- Manually verified in the running app.

### DIFF
--- a/rayforge/builtin_addons/rayforge-addon-laser/laser_essentials/steps/contour_step.py
+++ b/rayforge/builtin_addons/rayforge-addon-laser/laser_essentials/steps/contour_step.py
@@ -181,15 +181,6 @@ class ContourStep(Step):
         if not optimize:
             per_wp = [t for t in per_wp if t.get("name") != "Optimize"]
 
-        LeadInOutTransformer = transformer_registry.get("LeadInOutTransformer")
-        if LeadInOutTransformer:
-            calc = getattr(LeadInOutTransformer, "calculate_auto_distance")
-            auto_distance = calc(machine.max_cut_speed, machine.acceleration)
-            for t in per_wp:
-                if t.get("name") == "LeadInOutTransformer":
-                    t["lead_in_mm"] = auto_distance
-                    t["lead_out_mm"] = auto_distance
-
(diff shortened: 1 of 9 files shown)


knipknap:
Wow, thanks, good fixes!

PR #322, opened 2026-07-30
fix(mac): restore Monterey-compatible bundles

pgilfernandez:
This PR restores macOS Monterey compatibility for Rayforge bundles and hardens the macOS packaging process.

## Root cause

SciPy was upgraded from 1.11.4 to 1.17.1 in the macOS bundle. Recent SciPy wheels use Apple's newer Accelerate LAPACK interface, which requires macOS 13.3 or later.

On macOS Monterey, Rayforge failed during startup with:

```text
Symbol not found: (_dstevr$NEWLAPACK)
Expected in: Accelerate.framework
```

The startup exception was swallowed by the GTK application loop, causing Rayforge to exit with code `0` and incorrectly report that another instance was probably running.

## Changes

- Restore the macOS bundle pins:
  - NumPy 1.26.4
  - SciPy 1.11.4
- Require binary NumPy and SciPy wheels during macOS builds.
- Verify the installed versions and import the SciPy modules used by Rayforge before packaging.
- Set `MACOSX_DEPLOYMENT_TARGET` and `LSMinimumSystemVersion` to macOS 12.0.
- Return a non-zero exit code when startup fails before creating a window.
- Avoid reporting startup failures as an already-running instance.
- Fix the iterative Homebrew dependency-copying loop.
- Inspect the actual Mach-O launcher instead of the shell wrapper.
- Fix incorrect relative symlinks for bundled OpenCV libraries.
- Detect when `actool` succeeds without producing `Assets.car`, as observed on Monterey, and fall back to the legacy `.icns` icon.
- Use the virtual environment's Python consistently during packaging.
- Clean temporary requirement files when the build fails.

## Validation

Tested locally with:

- macOS Monterey 12.7.6
- Intel `x86_64`
- Python 3.11.14
- Rayforge 1.9.0-beta1

## Apple Silicon

The packaging changes are shared by the Intel and Apple Silicon builds.

Local validation was performed on Intel Monterey. The ARM and universal bundles should also be verified by CI.

### DIFF
--- a/Rayforge.spec
+++ b/Rayforge.spec
@@ -76,5 +76,8 @@ app = BUNDLE(
     name='Rayforge.app',
     icon=_icon,
     bundle_identifier='org.rayforge.rayforge',
-    info_plist={'CFBundleIconName': 'rayforge'} if _use_car else {},
+    info_plist={
+        **({'CFBundleIconName': 'rayforge'} if _use_car else {}),
+        'LSMinimumSystemVersion': '12.0',
+    },
 )
--- a/rayforge/app.py
+++ b/rayforge/app.py
@@ -28,6 +28,7 @@ from rayforge.logging_setup import setup_logging
 # ===================================================================
 
 logger = logging.getLogger(__name__)
+_unhandled_exception = False
 
 
 # Suppress NumPy longdouble UserWarning when run under mingw on Windows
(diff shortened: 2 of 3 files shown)


knipknap:
Thanks for the PR!

PR #323, opened 2026-07-30
Fix ChArUco detection array shapes

trixdaddy:
## Summary

- normalize ChArUco corner arrays to `(N, 2)` and ID arrays to `(N,)`
- support both singleton-dimension and flat OpenCV return shapes
- reject malformed, mismatched, empty, or insufficient detections
- add regression tests for both shape families and installed OpenCV behavior

## Problem

OpenCV 5 can return ChArUco corners as `(N, 2)` and IDs as `(N,)`.
The previous code indexed every corner as `pt[0][0]` and every ID as
`i[0]`, which raises `IndexError: invalid index to scalar variable` for
those flat shapes. This repeatedly crashes the Lens Calibration preview
and prevents Capture Frame from working.

## Verification

- `pixi run format`
- `pixi run lint` — 0 errors, 0 warnings
- `pixi run test -- -q` — 4913 passed, 2 skipped, 148 deselected
- `pixi run wheel` — wheel and sdist built successfully

Real-camera verification used the project Pixi environment with OpenCV
5.0.0 and an OBSBOT Tiny 4K:

- live result shapes were corners `(24, 2)`, IDs `(24,)`
- calibration preview displayed 24 detected corners
- five Capture Frame activations incremented the frame count to 5
- calibration became available and completed with five frames

### DIFF
--- a/rayforge/camera/calibration/charuco.py
+++ b/rayforge/camera/calibration/charuco.py
@@ -177,13 +177,24 @@ class CharucoBoard:
         if charuco_corners is None or charuco_ids is None:
             return None
 
-        if len(charuco_corners) < 4:
+        try:
+            corners_array = np.asarray(
+                charuco_corners, dtype=np.float32
+            ).reshape(-1, 2)
+            ids_array = np.asarray(charuco_ids).reshape(-1)
+        except (TypeError, ValueError) as error:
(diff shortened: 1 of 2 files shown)


knipknap:
Great, Thank you for the patch!

PR #344, opened 2026-08-12
fix: step name change not propagating to step list

### LINKED ISSUE #343: Step naming appears to be broken
Changing the name of a step appears to be broken.  Layer naming appears to work fine.

Problem example:

You create a new sketch in a layer, and name it something - for example, "Border Art".  It shows up in the layer dialog with the correct step name.  If you later double click on the step to edit it's properties, and change the name to something different, that change doesn't appear in the layer dialog when finished.  Instead, it retains the original name in the layer dialog.  Interestingly enough, if you double click on the step to edit it again, it shows the changed name that you had made before - but it still only shows the original name in the layer dialog.  Seems that the name change is not being propagated correctly back to the layer dialog for some reason, but is correctly being saved and displayed in the step properties when editing.

knipknap:
## Summary

Fixes the issue where renaming a step in the step settings dialog didn't update the step name shown in the main window's step list (WorkflowView).

**Root cause:** Two issues prevented the name change from propagating:

1. **`Step.set_name()`** did not explicitly fire the `updated` signal after changing the name, unlike `Layer.set_name()` which does. While the inherited `name` property setter does fire `updated`, making the notification explicit ensures consistency with the layer rename path and reliably notifies all listeners.

2. **`WorkflowView`** (which displays the step list in the right panel) listened to `workflow.updated`, `descendant_added`, and `descendant_removed`, but was missing `descendant_updated`. When a step is renamed, the change bubbles up as `descendant_updated` — which WorkflowView never received, so it never refreshed its display.

Added regression tests verifying that `set_name` fires `updated` and that the signal propagates to the workflow's `descendant_updated`.

Closes #343

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,4 +43,4 @@ external/
 
 # macOS icon build artifacts
 Assets.car
-rayforge.icns
+rayforge.icns.hermes_last_pull
--- a/rayforge/core/step.py
+++ b/rayforge/core/step.py
@@ -649,8 +649,10 @@ class Step(DocItem, ABC):
             self.updated.send(self)
 
     def set_name(self, name: str):
+        """Sets the step name and notifies listeners of the change."""
         if self.name != name:
             self.name = name
+            self.updated.send(self)
 
     def set_visible(self, visible: bool):
         if self.visible != visible:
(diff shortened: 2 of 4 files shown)


knipknap:
Merged separately with 0e4081213d83b83a6dfe065938b26d4caa2b280d.

PR #346, opened 2026-08-13
Add Amharic language support

yosephfeleke251:
This PR adds the Amharic language code mapping to the general preferences UI page and initializes the base translation file directory structure under locale/am/LC_MESSAGES/.


knipknap:
Hi there, thank you for opening the PR. I see that the file has no actual translations though, I think those would be needed before merging this. I am also wondering about maintenance - without frequent updates I expect this will be outdated quickly. Are you willing to take on maintainership for this language?

yosephfeleke251:
Hi @knipknap  thanks for the feedback!

​I am actively working on the translation now and will be pushing updates to this PR as I progress.

​I am adding Amharic support primarily to help my employees operate our laser cutter for paper card production. Because of this specific focus, I plan to translate the core UI they’ll come across daily, but I likely won't be translating features outside that workflow.

I completely understand if you'd prefer not to merge it without full coverage and ongoing maintainership, but I'm happy to provide these core strings if you're open to having a starting point for others to build on!

What do you think?

knipknap:
Ok, I guess it does not hurt even if this does turn out to be outdated. Thank you for the effort!

yosephfeleke251:
@knipknap Thank you! I appreciate it!

I plan to submit the translations in small batches as I work through them. Would it be okay to merge the initial batches along the way? That will help me test subsequent translations more easily.

knipknap:
Perhaps you can use the attached AI translation as a basis?

[rayforge.zip](https://github.com/user-attachments/files/31066403/rayforge.zip)

yosephfeleke251:
@knipknap Thanks for setting this up! Having this complete base file as a reference is going to save me a lot of time.

​AI often struggles with Amharic fluency, so there are bound to be inconsistencies and mistakes with technical terms, but it will be great to build on. I'll go through it, refine the core UI strings for our workflow, and push the updates.

​Out of curiosity, which tool did you use to generate the translations?

knipknap:
Great, Thank you! I used aistudio.google.com with Gemini 3.1 Pro in this case - it does not allow uploading .po files, but I just copy/pasted the content in the chat box. It is too large for it to respond with everything in one response, but I just asked it to continue in a second prompt and then combined the two responses in one file.

yosephfeleke251:
Ah, gotcha! Thanks for explaining the workflow.

​I'll dive in and push the updates to the PR soon.

PR #404, opened 2026-09-12
Beta3 - First PR for Ruida.

StevenIsaacs:
This is the first PR to add the capability to use Ruida controller based CO2 lasers.

The README.md in `rayforge/machine/driver/ruidarpa` contains more detailed information regarding this PR.

NOTE: This has been tested only with an RDC6442S based machine. Other controller variations may work but that can't be proven due to lack of hardware availability.

This implementation allows easy switching between using a direct connection or via the TUI (RPC). When using the TUI it must first be installed per instructions in the README. The TUI is included in this release because of the diagnostic and experimentation capabilities it provides. When ever possible use the TUI to understand unexpected behavior or to experiment with protocol features.

For verification, golden files are provided at `rayforge/machine/driver/ruidarpa/golden`. These files include Rayforge job files along with the resulting `gluescript` (.cglu), `rpascript` (.rds) , HTML plots and the RDWorks compatible files (.rd). The intent is to use these files to verify known behavior when diagnosing problems.

Features which should work reliably:
- VECTOR based designs.
- RASTER fill designs.
- DITHERed images.

Features which may work:
- Greyscale images: Per pixel power settings have been verified but actual results may vary widely.
- Depthmap images: These have not been tested and are currently not recommended due to the risk of using the Z axis.

Known problems:
-  Currently, any changes to the machine configuration causes reconnects with the Ruida controller. This is particularly annoying when entering numeric values in fields one keystroke at a time. This frequent reconnect can confuse the status monitoring layer. Recovery is simple. To re-establish comms toggle the TUI. Doing so will switch between using the direct driver and the RPC base driver.
- Greyscale images are not producing engraves with fine gradation between pixels. It is currently unknown how to best compensate for this. It may be this is a limitation of using a CO2 laser.

And of course, there is the common unknown list of unknown problems.


.vscode/launch.json
knipknap, line 1: I would commit this separately

.vscode/settings.json
knipknap, line 1: I think this should keep pointing to the default environment. I am not actually sure we need a dedicated Ruida environment

AGENTS.md
```diff
@@ -61,3 +61,6 @@ wrapper (or any other pixi command).
 - Do not make changes unrelated to the current task
 - Never remove logging or debugging unless asked by the user
 - Do not repeat files unless they have changes
+
+## Addendums
+- Read the AGENTS.md located in the ruidarpa driver directory at rayforge/machine/driver/ruidarpa.
```
knipknap: Maybe something like "When working on the Ruida driver, read the AGENTS.md..." to avoid filling the context unconditionally

pixi.toml
knipknap, line 1: I would actually not create a dedicated Ruida environment. The dependency seems small; the Ruida PA package could be made available in default

tests/machine/driver/ruidarpa/golden/regen_golden.py
```diff
@@ -0,0 +1,160 @@
+"""Regenerate the ruidarpa golden fixture.
+
+Run from the repo root inside the ruidarpa pixi environment:
+
+    pixi run -e ruida-pa python \\
+        tests/machine/driver/ruidarpa/golden/regen_golden.py
+
+The fixture locks the encoder's GlueScript transcript byte-for-byte
+(the ``staged_job.rpas`` filename is a historical misnomer — the fixture
```
knipknap: Maybe it can be renamed now to avoid renaming it later?

PR #416, opened 2026-09-14
Multi machine settings dialogues open at the same time.

StevenIsaacs:
# rayforge 1.11.1 Testing Log

### Problem

Clicking Machine → Machine Settings, the maintenance alert's "View Counters"
action, or the machine warning banner each constructed and presented a fresh
`MachineSettingsDialog`, so multiple Machine Settings windows could be open at
the same time. There was no instance tracking.

### Solution

`MainWindow` now tracks a single dialog instance in a `_machine_settings_dialog`
field. All three entry points route through a new `_present_machine_settings_dialog()`
helper that re-presents the existing dialog when it is still bound to the active
machine, auto-closes a stale dialog after a machine switch, and otherwise creates
and stores one dialog. A `close-request` handler resets the tracked reference when
the dialog closes. `MachineSettingsDialog` gained a `select_page()` method so
re-presenting an existing dialog can still navigate to the requested page.

### Verification

Two new UI tests in `tests/ui_gtk/test_mainwindow.py`:
`test_machine_settings_dialog_is_single_instance` (opening twice yields the same
instance; closing clears the tracked reference; reopening yields a new instance)
and `test_machine_settings_dialog_recreated_after_machine_switch` (switching the
active machine invalidates the open dialog and a fresh one is created bound to the
new machine). Both pass via
`pixi run --environment default pytest -v -m ui tests/ui_gtk/test_mainwindow.py -k machine_settings`.
`pixi run lint` reports no new findings for the modified files (the 18 pyright
errors are pre-existing in unrelated files).

### Problem

The maintenance alert's "View Counters" action passed `initial_page="hours"`, but
no sidebar page is named "hours" — the counters live on the "maintenance" page, so
the dialog opened on the default page instead of the counters.

### Solution

`_open_machine_hours_dialog` now passes `initial_page="maintenance"`, so the dialog
opens directly on the page containing the counters.

### Verification

The change is a one-line argument correction routed through the shared
`_present_machine_settings_dialog()` helper; it compiles cleanly and `pixi run lint`
reports no new findings.


knipknap:
Thanks for the PR!

PR #423, opened 2026-09-17
Add pointer laser offset with work-zero compensation

knipknap:
## Summary

Adds a LightBurn-style **pointer laser offset**: users with a separate pointer (red dot) laser mounted at a fixed distance from the cutting beam can enter that offset per laser head and toggle the compensation on/off. This removes the manual XY-offset arithmetic needed when aligning stock by the pointer dot — especially for pass-through work where stock is larger than the bed.

## Behavior

When the offset is enabled for a head:

- **Set Work Zero at Current Position** and the **Zero X / Zero Y** buttons add the offset, so the work origin lands where the *pointer dot* marks the stock instead of where the (hidden) cutting beam is.
- The **canvas laser dot** is drawn at the pointer dot's position, matching what the user physically sees on the material.

Deliberately unchanged: cutting, framing, jog, and "Move to" commands always refer to the cutting beam, so **G-code output is bit-identical whether the offset is on or off**. Click-canvas-to-set-zero also stays in design coordinates.

## Implementation

- `LaserHead` (`machine/models/laser.py`): new `pointer_offset_enabled` / `pointer_offset_x` / `pointer_offset_y` reviewable settings, serialized into the machine profile; defaults (off, 0/0) are fully backward compatible with existing profiles.
- `Machine.get_pointer_offset(head=None)` (`machine/models/machine.py`): resolves the offset for a given or the default laser head, returning `(0, 0)` when disabled so callers can add it unconditionally.
- `MachineController.set_work_origin_here` (`machine/models/controller.py`): applies the X/Y compensation when writing the active WCS slot — a one-time bake into the offset, so all downstream WCS math (canvas origin, job pipeline, frame) stays consistent.
- `WorkSurface.set_laser_dot_position` (`ui_gtk/canvas2d/surface.py`): offsets the dot in machine space before the WCS-aware canvas mapping; re-renders immediately when the setting changes.
- **Machine Setup → Heads**: new "Pointer Offset" group with an enable switch and X/Y mm fields (fields gray out while disabled).

## Docs

- `machine/laser.md`: Pointer Offset section incl. a measuring recipe (pointer mark vs. focus-power beam mark → DRO delta) and a rotary-mode caveat (Y offset not meaningful).
- `features/workpiece-positioning.md`: Pointer Laser Offset section plus a step-by-step pass-through workflow.

## Test plan

- [x] `tests/machine/models/test_laser.py` — defaults, setters/signals, serialization round-trip, missing-field backward compat, `Machine.get_pointer_offset` resolution
- [x] `tests/machine/models/test_machine.py` — work-zero compensation with offset enabled (265 tests pass in file)
- [x] `tests/ui_gtk/canvas2d/test_laser_dot_position.py` — dot follows pointer position; disabled offset leaves dot at beam position
- [x] `pixi run lint` clean (pyflakes, pyright, formatting)


AVRASM2:
Hi, i just tried it out, looks promising :+1: 

i just have issues to work precisely or my UX is different:

1. RF doesn't support "Snap object to Laser-Dot" kind of things yet (Corners / Centers [9-Points] )
(MeerK40t did, but it has lots of other unpleasant bugs & lack of features & UI/UX)
so i couldn't easily Align Laser, then snap object to it, do 1-2 straight moves to confirm rotation being 0°

2. that way as you planned, i can't precisely move the Aiming-Laser to the corners or Frame around to confirm it will be exactly where i need it to double-check (i still need to enter G-Codes or bodge a Macro to move the Head - that is not beginner-friendly at all tbh!)

**0.1-0.25mm is possible by eye with a fine & dim spot**

it makes somehow sense to not mess with the positions as you intended  
but actually it would be nice to have a **Button** somewhere in that control-panel (i.e. right to the WCS selector)  that somehow applies that Offset to "Move-To Corners/Center" and "Framing" (and logically Jogging)  

pressing "Start" ~automatically reverts that Button off course~ **(see edit below!)**
Toggling that Button directly moves the Machine and a Yellow Dot could be displayed instead, and the Button should have 2 distinct colours, like yellow or orange (depending dark / light theme) if active, and regular gray or what ever if OFF

actually you could show a Red & Yellow Dot & Circle together? no clue what would make sense UI/UX wise but somehow that the User sees where it Fires & where it Aims, and also that he can select which one to use to do Aligning (that Button)
the active one is a full Dot, the inactive a hollow circle - that makes sense to me

if Framing / Move-To would be outside the Machine-Bed, simply give a warning and move to the max. possible location, to still be able to see "straight on machine?" or so  
(but most jobs are way smaller than the Laser, especially those that requires highest precision alignment! where i would require the Aiming-Laser to be where other wise only the Burning-Laser would fire to have best precision and confidence i won't destroy a "only 1 chance Job" by 0.25-1mm offset or worse) 

the WCS feature somehow doesn't work like on my CNC with gSender, so i couldn't simply have G54 = Real Laser, G55 = Aiming Laser (or i do something entirely wrong with it / don't understand yet how it actually works here)
otherwise i would first enter G55 values, then switch to G55 and move the Aim-Laser around, switch back to G54 and run it with the Burning-Laser
(essentially what that Button would do somehow)

maybe you & other users have also some Inputs to my feedback to fine-tune that rather important feature? 

best Regards :smile: 


ps:
Move-To doesn't honor the speed entered in the Jog-UI, but uses F1500 instead, only if i alter the G-Code settings-string to use a different hard-coded F rate instead of **F{speed}**, it always is that slow!

ps: (edit)
i realized i would also like to run the entire job on the Auxiliary Aiming-Laser as well, so hitting "Start" should NOT revert that offset automatically **but give a Warning and let the user choose "Burn", "Alignment", "Cancel"**
that way anyone can simulate his job in real life and watch for other mishaps like "i positioned that one Text at the wrong spot" on his precious 1-chance piece :smile: 
as a safety: if "offset" is active, it should force the S value to "Frame Power", or maybe entirely to 0% ?

that would result into the same workflow as with a Fiber-Galvo laser that actually can switch between Red & Fire, and scan a preview of what it will laser and where  
(i had access to a Trotec SpeedMarker MOPA - really expensive gear - but it sadly had an Offset and no clue how to callibrate it out, except having to move the job in their software before actually starting it)

good case are PCBs:
first i engrave into the Copper (usually those with photoresist or lacquer so i can finish it by etching, not lasering into the base-material at all, only laser away max. 3/4 of the copper)
then i apply solder-resist across the entire surface, harden it without photomask evenly
then i need to precisely (0.1-0.2mm tolerance) re-align it in the Laser against etched marks, and open the solderpads up again, some parts have 0.5mm pitch...

often i can ease it with a fixed corner between those 2 laserings, but not always (i.e. it lost position or gantry shifted)

Issue #428, opened 2026-09-18
Flaky behavior on Windows 


knipknap:
Kandilyabir posted the following on Discord:

I am posting the information here.
As promised, I have attached the saved program log from Windows 10 (please let me know if further details are needed) and a screenshot of the program window.
At some point, the program stopped responding correctly to the machine control buttons. The program itself did not close. As shown in the screenshot, the command button icons turned gray, even though they had worked initially and I had successfully performed parking and workspace tracing operations several times.
During the workspace trace, I didn't know how to stop the process, so I used the "Cancel Running Job" button. Movement stopped, but the "Clear Machine Alarm" icon became active. In some instances, clicking this icon caused the engraving area trace to resume (which, it seems to me, shouldn't happen).

I was unable to manually turn on the laser to check the focus: pressing the button had no effect, even when adjusting the minimum laser power. It is possible I simply didn't figure out the correct procedure...

![2026-09-18_205806.png](https://github.com/user-attachments/assets/a2f2b605-81af-4470-a1ad-a9b8293c5c46)

![rayforge_debug_2026-09-18_17-56-52.zip](https://github.com/user-attachments/files/32393611/rayforge_debug_2026-09-18_17-56-52.zip)

----

A quick update:
After restarting the program, I noticed that it was no longer behaving as it had during the initial launch. Specifically, the "test fire" icon in the lower laser control section is currently inactive, whereas it was active—along with the icon at the top—when I first started the program.
Additionally, once I initiated the workspace outline/framing function, I was unable to stop it because the control icons at the top remained inactive.

I'll attach the log as well, just in case.

![2026-09-18_212233.webp](https://github.com/user-attachments/assets/24998f31-05f1-4433-b78f-1a537d41ba05)

![rayforge_debug_2026-09-18_18-25-38.zip](https://github.com/user-attachments/files/32393634/rayforge_debug_2026-09-18_18-25-38.zip)

------

This time, the program closed after all...

![session-2026-09-18_18-36-34-1.log.txt](https://github.com/user-attachments/files/32393700/session-2026-09-18_18-36-34-1.log.txt)

knipknap:
A correction to my earlier analysis: I wrote it too quickly and several details were wrong. I have now gone through all three of your logs line by line — the story is different (and more interesting) than I first described, and the mystery from your second screenshot is explained too. My apologies for the confusion; here is the corrected version.

**What actually happened when the trace resumed**

[...]

PR #432, opened 2026-09-19
fix(grbl): heal flow-control state jammed by lost acks (issue #428)

knipknap:
## Problem

Follow-up to #428 / PR #429, found in the 1.11.2 logs ([comment](https://github.com/barebaric/rayforge/issues/428#issuecomment-5743515226)).

After cancelling a job, the driver sends its safety shutdown (`M5 S0 G0`, `M5`, `M9`) to the freshly soft-reset controller. On the Sculpfun iCube over Bluetooth, the firmware acknowledges only the first of these — the other acks never arrive. Their pending entries then poison the flow-control queue permanently:

- every later `ok` (e.g. for `$X`) is mis-attributed to a stale FIFO entry, so the accounting never returns to zero, and
- `execute_interactive_command()` starts with an **unbounded** `pending_queue.join()`, so it deadlocks — "Read from Device" hangs, the UI stays busy/inactive while status polls keep flowing. In the log the first read only "finished" because the user clicked Read again, which replaced the stuck task; the queue itself stayed poisoned (`buf: 8/511` for minutes).

## Changes

- `GrblSerialTransport` timestamps the oldest un-acked command and exposes `oldest_pending_age()`.
- `execute_interactive_command()` drains via `_await_pending_drained()`, which heals (resets) the flow-control state once the oldest pending entry outlives `STALE_PENDING_ACK_TIMEOUT` (10 s) instead of deadlocking.
- The idle connection loop proactively heals stale pending state each poll cycle, so the queue unjams within seconds even without user interaction.
- Healing never runs while a job is streaming (`_job_running`), where un-acked entries age legitimately during long moves.

## Tests

Three regression tests in `TestIssue428StaleAckJam`:

1. `test_interactive_command_survives_lost_ack` — a 60 s-old un-acked entry no longer deadlocks `execute_interactive_command` (fails/hangs pre-fix).
2. `test_connection_loop_heals_stale_pending` — the idle poll loop clears stale entries on its own.
3. `test_stale_ack_not_healed_during_job` — legit in-job pending entries are left alone.

`tests/machine/` passes (2359 tests).


knipknap:
Good call — the flat 10 s threshold would have treated a slow move (or homing) as a lost ack. Reworked in 247e4b106:

- `PendingCommand` now carries its **own ack deadline**: `send_gcode()` stamps each entry with the estimate-derived per-line timeout it already receives (`max(5 s, estimate × 3)`, capped at 120 s — the exact values `5fb7f9e82` introduced for stall handling). Interactive/queue commands keep a short default (10 s, matching the existing interactive-command cap in `_execute_command`).
- The transport reports `pending_ack_overdue()` only when the oldest entry outlives **its own** deadline. When the queue head changes, the stamp is refreshed conservatively, so healing can only happen later than an entry's true deadline, never earlier.
- Healing still never runs while a job is streaming (`_job_running` gate) — during streaming the stall machinery owns recovery, as before.

New regression test `test_slow_command_not_healed_within_deadline` pins this: an entry with a 120 s deadline is not healed even at 60 s of age, which the previous fixed threshold would have wiped.

PR #434, opened 2026-09-19
fix(svg): import viewBox-only SVGs without width/height

MausRundung:
## Problem

SVGs that carry only a `viewBox` and no explicit `width`/`height` attributes (common Illustrator/Inkscape exports) fail to import with:

> Could not determine valid SVG dimensions.

The import dialog also shows no natural size for such files, so they are effectively undetectable/unclickable.

## Root cause

A regression introduced when SVG metadata parsing migrated from `svgelements` to `raygeo.svg.extract_svg_metadata`. `raygeo` returns `None` for `width`/`height` when the file has no explicit attributes — even when a valid `viewBox` is present — whereas the old code path fell back to the `viewBox` size.

## Fix

Add a `viewBox` fallback in the two places that matter, interpreting the viewBox user units as pixels (the SVG spec default) at the requested PPI:

- `rayforge/image/svg/svg_base.py` — `_get_svg_parsing_facts()` falls back to `meta.viewbox` width/height when pixel dimensions are `None`, so the importer no longer aborts.
- `rayforge/image/svg/svgutil.py` — `get_natural_size()` derives the mm size from the viewBox, so the import dialog shows the correct dimensions.

`is_unitless_svg()` already returns `True` for these files, so the DPI control is presented as expected.

## Tests

Added `test_get_size_viewbox_only` (a 475×475 viewBox-only SVG, mirroring the real-world file that motivated this) asserting `get_natural_size()` resolves to `475 * MM_PER_PX` on both axes. Full `tests/image/svg/` suite passes.


knipknap:
Thank you for the fix!

PR #435, opened 2026-09-19
fix(camera): crash-free device switching and non-blocking picker

MausRundung:
## Summary

Fixes the camera stack so that switching devices no longer crashes the app or freezes the picker. Grouped as a sequence of independent commits (each can be reviewed/backed out on its own).

### The crash (device swap)

The old device-change path orphaned a capture thread and then rebuilt the controller, so two threads opened the same device (`cv2.VideoCapture` double-open) — the classic segfault on Windows DSHOW/MSMF.

- **serialize stream start/stop** — `_start_capture_stream`/`_stop_capture_stream` now run under a lifecycle lock; `VideoCaptureDevice` takes a cancel event so open-retry/reconnect/frame-pacing waits abort promptly on stop; the running `cap` is force-released if the thread survives the join, so `join()` actually succeeds.
- **swap devices in place** — the controller now owns the device it opened (`_device_id`) and the capture loop reads that instead of `config.device_id`, so a live model mutation can't make a running thread retarget. `set_device_id()` stops/restarts cleanly; `dispose()` disconnects the config signals so a destroyed controller can't be resurrected. Only true add/remove hits controller teardown.
- **guard frame buffers** — `_image_data`/`_raw_image_data`/`_accumulator` are published and snapshotted under a lock, so UI-thread readers never observe torn/`None` state during a swap.

### The freeze (picker)

- **non-blocking picker** — `CameraSelectionDialog` opens instantly with a spinner; the device scan and previews run on a worker thread and marshal back via `GLib.idle_add`. The scan is cached briefly, and devices with a live controller are previewed from its last frame instead of a throwaway open — no more double-open of the running camera and no more "Not Responding" window.
- **GTK4 spinner** — `set_spinning(True)` instead of the GTK3 `spin()` (pyright/CI fix).

### Tests

- The `list_available_devices()` int-device-id regression now targets the in-process scan path (`_scan_cameras_fallback`), since the real call probes in a `spawn` subprocess where in-process mocks can't cross the boundary. Mock gained a `read()` for the frame-validation step.

## Commits

- WIP: frame validation, subprocess scan, device-swap UI
- serialize stream start/stop and force-release wedged captures
- swap devices in place and detach destroyed controllers
- make the camera picker non-blocking and stop double-opening
- guard frame buffers with a lock
- satisfy ruff SIM102/BLE001 in device-swap fixes
- use GTK4 Spinner.set_spinning() instead of spin()
- retarget int-device-id regression test to in-process scan

## Notes

This branch also carries a small `.vscode/settings.json` change that came along with the initial WIP commit — happy to drop it from the PR if preferred.


.vscode/settings.json
MausRundung: I hate it when an IDE clutters Tree views... didn’t mean to commit this (shouldn’t matter for your end tho)

MausRundung:
This commit is partly 3 months old. Its main purpose is to stop Rayforge from crashing when you have more than one camera on Windows, but it also allows you to swap/fix/change a camera if Windows or Rayforge messed up camera IDs again, so you don't have to reconfigure it entirely.

MausRundung:
This also will probably be replaced once the system of "Kaper" is tested and fully Implemented but till then this will keep it working

knipknap:
@atkaper any thoughts on this PR? Given its size, I worry about tripping up your work if I merge it. Perhaps it is better to merge this into your changes? What do you prefer?

atkaper:
@knipknap I will have a look, thanks for holding off for a bit.

atkaper:
Note: https://github.com/barebaric/rayforge/pull/438 if deemed suitable would replace this PR.

knipknap:
@MausRundung what do you think, does the new PR fully address the issues? If it does, can your PR be closed?

MausRundung:
I would say yes but once the big one is merged i would do some changes to Noise and Camera Settings in general: but this PR can be closed.

knipknap:
Ok, Thank you for the PR anyway!

Issue #422, opened 2026-09-16
Option to choose M3


sms043:
Add option to use M3 on constant power situations. 
When using M4 sometimes the laser power falls below the required power on corners. 
Changing the command template "Laser On" to use M3 fixed it. 
thanks

knipknap:
Thanks for the proposal - though changing the power command like you did is exactly the intended way, so I am not sure what else could be done here.

sms043:
thanks

AVRASM2:
Hello, i found out you have a "Focus Laser On" command in the settings, it uses "M3 S{power:.0f} G1 F1"   
**(G1 F1 i've added otherwise my IKIER won't turn it on!)**  

for **Framing operation** it could make sense to use that setting as well, as currently it uses "Laser On" which is "M4 S{power:.0f}" and that is useless in framing, where the Laser needs constant power regardless of moving speed  ?
  
btw, Framing should also work at 0% power! i.e. for some Lasers with Auxiliary Alignment-Lasers, just somehow please add a Button to select & move to that Offset and back (for Jogging & Framing, when Start: cancel automatically and move back)  that can be entered in the Settings would be a great Feature :-)
i.e. my FiberHead has an Offset of where it "Aligns" (Red Dots) and where it actually "Fires" (IR-Laser)

**sorry to Hi-Jack this thread, but somehow it is actually related to M3 / M4 thematic...**

my 2 cents to the original post:
maybe it could be useful for some recipes  to be able to force M3 for certain operations while leaving M4 for others active and vice versa?
especially for mixed Workflows on 1 Piece somehow?
i.e. "engrave with M4, Cut with M3" ?
like a "Post-Processing"-Option that allows to choose/override or so and also save in a recipe for later

best regards :-)

knipknap:
Thanks for the report! This is a known trade-off of M4 (dynamic power mode): the firmware scales laser power proportionally to head speed, which is ideal for raster engraving, but on some machines the power dips below what is needed at corners and curves where the head decelerates. Constant power (M3) avoids that — and until now the only way to get it was to hand-edit the **Laser On** template in the G-code dialect editor, as you found. Far too obscure for something this common, so we're making it a proper, first-class process setting.

**Planned architecture**

[...]

Issue #35, opened 2025-09-27
Allow Baud rate to be edited manually -OR- update to add 1,000,000


loneregister:
**Is your feature request related to a problem? Please describe.**
I am unable to connect via USB to my MonPort Mega 70W laser. The documentation from MonPort specifies that the baud rate should be 1,000,000.  I am able to manually set this in lightburn, which then works.

**Describe the solution you'd like**
Add 1,000,000 to the drop down (quick). Or allow manual entry of the baud rates. Or allow drop down usage + manual entry of baud rate.

**Describe alternatives you've considered**
I am not aware of an alternative.

**Additional context**
None

knipknap:
I assume that this was done by #36, can I close it?

loneregister:
Resolved by 36.  Closing

PR #437, opened 2026-09-20
feat(laser): per-step power mode (constant M3 / dynamic M4)

knipknap:
Implements the per-step constant/dynamic power choice enabled by raygeo 1.53.x (barebaric/raygeo#34, #35), resolving #422.

## What

**Power Mode step setting** (`LaserStep`, inherited by Contour, Engrave, Frame, etc.):
- New "Power Mode" choice in the Laser settings group: `Dynamic (M4)` (default) / `Constant (M3)`.
- Wired into the assembler job path: `populate_payload` sets `ComputePayload.power_mode` (raygeo 1.53.1), which the compute stage injects as a `SetPowerMode` state op — so the setting actually reaches the G-code. (An earlier revision only stamped `create_initial_ops`, which the assembler pipeline doesn't consume; caught by community testing in #422.)
- Included in serialization, recipe keys, cache params, and `get_settings()`. Existing documents keep `DYNAMIC` and produce identical output.

**Encoder**: `dialect_to_spec` now passes `focus_laser_on` through to the raygeo encoder (previously driver-only), which emits the dialect's constant-power template (`M3 S…`) instead of `laser_on` (`M4 S…`) when a step requests constant power. Dialects stay pure syntax; no template changes.

**Framing**: frame ops always set constant power. On stock Grbl with laser mode enabled (`$32=1`), a stationary M4 emits no beam at all, so M4 framing was fragile; M3 frames reliably at any controller.

**Framing at 0% power**: zero frame power previously cancelled the frame action and disabled the frame button. Zero is now valid — the frame traces the outline with the beam off (the encoder omits the laser-on command at 0% power), so machines with an auxiliary alignment laser can frame without firing the main laser. `Machine.can_frame()` now requires any laser head instead of non-zero frame power; the Frame Power row description no longer calls 0% "disabled".

**Laser tab**: `LaserSettingsPage` renders the new Power Mode row (its key list is explicit, so the var alone wasn't enough).

## Mixed workflows

Both modes can coexist in one document — e.g. engrave a raster with dynamic power, cut the outline with constant power — since the choice lives per step on the op stream, not on the machine or dialect.

## Dependencies

- Requires **raygeo 1.53.3** (1.53.1 adds `ComputePayload.power_mode`; 1.53.2 preserves it through Optimize; 1.53.3 gives `PowerMode` a proper `__str__` for choice-widget round-trips).

## Tests

- New `LaserStep` tests: defaults, setter validation/signals, raygeo mapping, `create_initial_ops` op emission, **payload wiring**, and an **end-to-end regression test** (step setting → `ComputePayload` → raygeo pipeline execution → `SetPowerMode` in the ops) covering the real job path.
- Serialization round-trip; legacy docs default to dynamic.
- New encoder tests: constant mode emits the focus template, dynamic stays on `laser_on`.
- New framing test: 0% power traces the outline with no `M3 S`/`M4 S` commands; framing regression test updated to expect 12 × `M3 S` laser-ons.
- Laser page UI test asserts the Power Mode row renders on the Laser tab.
- `tests/pipeline`, `tests/machine/models`, laser addon, `tests/core`: 2277 passed. `pixi run format` + `pixi run lint` clean; laser page UI suite green under xvfb.


AVRASM2:
Please See: https://github.com/barebaric/rayforge/issues/422#issuecomment-5751605481

sms043:
I tried #437. I found the laser power mode setting in contour. I changed it to constant power mode. The gcode did not change, still M4. I found the gcode template for laser focus and changed it from M4 to M3 but the gcode for the contour did not change.
[rayforge_debug_2026-09-20_19-46-07.zip](https://github.com/user-attachments/files/32443238/rayforge_debug_2026-09-20_19-46-07.zip)

knipknap:
@sms043 you commented the second I pushed a fox for this issue. Can you try again?

sms043:
Still used M4. Power Mode set to M3.
[rayforge_debug_2026-09-20_23-00-08.zip](https://github.com/user-attachments/files/32446576/rayforge_debug_2026-09-20_23-00-08.zip)


G21 ;Set units to mm
G90 ;Absolute positioning
G54
M8
T0
G0 X-60.599 Y-16.691
M4 S128
G1 X-49.463 F500 S127.5
G1 Y-16.877 S127.5
G1 X-50.669 Y-17.526 S127.5
G2 X-53.824 Y-19.381 I-19.306 J29.233
G1 X-54.845 Y-19.937 S127.5

sms043:
No luck. I uploaded debug file to github
thanks

knipknap:
Second wiring gap found and fixed, thanks @sms043 for the debug bundle and retest.

Your G-code pinpointed it: the contour step's default transformer chain includes **Optimize**, which rebuilds the op stream from state snapshots — and that rebuild re-emitted every state command *except* the new power-mode one. So the mode reached the assembler but died at optimization, and the encoder fell back to `M4`. Fixed in raygeo (v1.53.2, barebaric/raygeo#36) and pinned here; there's now a regression test that runs a constant-power step through assembler **+ Optimize**, which reproduces your exact failure before the fix.

Please re-fetch `pull/437/head` and try once more (`pixi run` will pull raygeo 1.53.2 — make sure it says 1.53.2 when the env syncs). With your dialect, Constant (M3) should now emit `M3 S128` before the first cut instead of `M4 S128`.

(If it still shows M4 after that, please attach another debug bundle and I'll keep digging — but both known drop points, payload and optimize, are now covered by end-to-end tests.)

sms043:
Yes that fixed it. The laser on gcode shows M3 S128.
My CO2 laser does not fire very well at low power setting caused by M4 in corners. This will fix it with constant power. 
thank you

knipknap:
Thank you again for testing!

PR #438, opened 2026-09-20
Camera refactor, add stream sources (network cameras)

atkaper:
WARNING: backup your machine config before testing this PR.

@knipknap / @MausRundung Sorry, this got quite out of hand ;-) On each test I found something which needed another "small" change, all adding up to quite a big PR. As mentioned else where, this PR has been coded using AI. Taking small steps each time, and checking results looked sensible to me. I am not a native python developer (but I am a senior developer), so make sure to do a thorough review.

I did test my PR manually as well, as having AI add tests. I have run this on Linux Mint XFCE, with a Creality Falcon A1 Pro (20W blue) attached. I am quite happy I now can see my camera feed ;-)

I did do a quick run in virtualbox windows 10, and in there I also see me camera, and a local usb connected one.
But maybe @MausRundung can do some tests in windows as well? (as I'm not a windows user).

If this PR looks suitable for you, I guess you can close https://github.com/barebaric/rayforge/pull/435 as I did add the thoughts and architectural changes from it in this PR as well.

Ok, so now for the (AI) PR description - hope it helps in your review:

Thanks for your time & Regards,
   Thijs.
---

# Camera source refactor PR notes

This document explains the camera refactor that introduces a stable camera
identity and support for multiple camera source types:

- local attached cameras
- HTTP/HTTPS snapshot URLs
- HTTP/HTTPS stream URLs
- RTSP streams

It is written as a handoff document for PR review and as source material for
future user-facing documentation.

The main motivation for this PR is support for my "Creality Falcon A1 Pro"
camera feed, which exposes an HTTP snapshot endpoint that returns one JPEG
image per request. The previous camera stack assumed that all cameras were
local devices, so using this feed required a custom patch. This refactor
supports multiple camera source types while keeping the existing
calibration/alignment/overlay pipeline intact.

> **Configuration backup and rollback note:** This branch writes both the new
> source-based camera configuration and a temporary legacy top-level
> `device_id` field for local cameras. The legacy field is intentionally kept
> so users can return to an older Rayforge version after testing this branch
> without the older version failing on a missing `device_id`. It is planned
> for removal after the migration has been established for a few months, so
> keep a configuration backup before testing. Cameras created with the new
> HTTP or RTSP source types are not usable by older versions; remove or
> manually clean up those entries before going back.

## Why the camera module was refactored

Before this refactor, the camera stack assumed that every camera was a local
device that could be opened directly through OpenCV `VideoCapture` using a
numeric device index or a Linux `/dev/v4l/by-id/...` path.

That worked for USB webcams, but it mixed together two concerns that should be
separate:

1. acquiring frames from a source
2. processing and using those frames inside Rayforge

This became a blocker for network-fed cameras because HTTP snapshot feeds,
HTTP MJPEG-like streams, and RTSP streams do not behave like local devices.

The main design goal of this PR is therefore:

> Keep the existing calibration/alignment/overlay pipeline intact, but insert
> a source abstraction layer underneath it so different camera transports can
> be plugged in cleanly.

## High-level design

The refactor separates camera identity, camera configuration, and camera frame
acquisition.

### 1. Stable camera identity

Each camera now has an internal UUID-backed `id`.

This ID is used for:

- `CameraManager` controller lookup
- keeping UI/controller relationships stable
- surviving edits to names or source details

This avoids using `device_id` as a global identity key. That old design was
not suitable for the new features because a source path or URL is
configuration, not identity.

### 2. Source type and source config

Each camera now stores:

- `source_type`
- `source_config`

Current source types:

- `local_device` (this is the previous USB/local-camera behavior)
- `http_snapshot` (this is for cameras that return one still image per request)
- `http_stream` (this is for cameras that return a continuous HTTP/HTTPS stream)
- `rtsp` (this is for cameras that return an RTSP stream)

Examples of `source_config`:

```json
{"device_id": "/dev/v4l/by-id/usb-Logitech_Webcam-video-index0"}
```

```json
{"uri": "https://camera.example.local/snapshot.jpg"}
```

```json
{"uri": "https://camera.example.local/mjpeg"}
```

```json
{"uri": "rtsp://camera.example.local/stream"}
```

### 3. Source abstraction under the controller

`CameraController` continues to own:

- stream lifecycle
- reconnect behavior
- denoise
- lens correction
- alignment transform
- pixbuf generation
- image delivery to UI/canvas consumers

Frame acquisition is delegated to source classes in
`rayforge/camera/source.py`.

This keeps the existing downstream camera pipeline reusable across all source
types.

### 4. Crash-free source switching and discovery

The local USB source switching implementation also incorporates architectural
ideas from `barebaric/rayforge` PR #435, while retaining this branch's
`CameraSource` abstraction and stable camera IDs.

The controller now:

- serializes stream start and stop with a lifecycle lock
- uses a cancellation event to interrupt source open retries, reconnect
  delays, and frame pacing
- waits for the old capture thread to terminate before starting another one
- refuses to start a replacement stream if the previous thread remains stuck,
  rather than risking a second hardware open
- keeps the source owned by the running capture loop separate from mutable
  camera configuration
- disconnects model signals when a controller is disposed
- publishes and reads raw/processed frame buffers under a frame lock

Local device sources register their device before opening it and discovery
does not probe devices already held by a live source. The picker performs
device discovery and one-shot previews asynchronously, reuses the latest
frame from an active controller, and ignores stale results after a dialog is
closed or a different device is selected.

Device discovery is also no longer triggered by unrelated configuration
changes. The camera properties panel previously re-scanned the hardware on
every model `changed` signal, so editing an unrelated value (for example a
lens distortion coefficient in the manual calibration dialog) opened and
released every local capture device on the UI thread. Repeating that quickly
made the UI stutter, flooded the log with backend probe warnings, and could
collide with a capture stream reopening its own device. The panel now tracks
which source its device list was built for and only re-scans when the source
type or device actually changes.

For the same reason, the distortion coefficient setters now ignore writes
that do not change the stored value, matching the existing behavior of the
brightness, contrast, and white balance setters. Previously every write
emitted change signals and re-stamped the calibration date, which also
invalidated the image alignment even when nothing had changed.

Windows camera backends can report `isOpened()` successfully before they can
deliver frames. Local-source opening and discovery therefore perform initial
frame validation and try the next backend when a backend cannot produce a
usable frame. This is particularly relevant for DirectShow and Media
Foundation camera drivers.

These changes address the device-switch crash and picker freezes without
changing the calibration data associated with the logical camera.

### 5. Balanced stream subscriptions in the camera UI

The controller runs its capture thread while at least one consumer is
subscribed, so every `subscribe()` must be matched by exactly one
`unsubscribe()`. The camera wizard violated that contract: pages
released their stream when navigated away from, and the wizard released
them a second time on close, while the capture page never released its
own. Cancelling the wizard therefore drove the subscriber count below
the wizard's share and stopped the stream for unrelated consumers, most
visibly the main canvas preview, which had no way to notice and
resubscribe. The feed stayed dead until restart.

The camera widgets that own a subscription (`CameraDisplay`, the
calibration capture surface, the alignment surface, and the image and
lens settings widgets) now expose idempotent `start()`/`stop()` pairs
that track whether they are currently streaming. Wizard pages acquire
the stream in `enter()` and release it in `leave()`, so navigating back
to a page also restores its live preview, and the wizard's `close()`
simply leaves every page instead of special-casing page types.

`CameraController.unsubscribe()` additionally logs a warning when the
subscriber count is already zero, so a future imbalance is visible in
the session log rather than only as a silently dead feed.

### 6. Lens calibration no longer terminates the wizard

Capturing Charuco frames only fills the calibrator's buffer; the lens
model is solved when the user presses **Calibrate** and confirms
**Save Calibration**. Two flow defects made that result unreachable in
practice. Saving closed the entire wizard, so lens calibration and
image/world alignment could never both be completed in one pass. And
the capture page placed no condition on **Next**, so walking straight
through to alignment silently discarded every captured frame and left
the camera uncalibrated with no indication that anything was lost.

Saving a calibration now advances to the following page (alignment)
and confirms with a toast, and the capture page vetoes **Next** with a
confirmation dialog when frames are captured but not yet solved.
Pages can express that veto through the new
`CameraWizardPage.on_next_requested()` hook, and resume navigation via
`CameraWizard.advance()`.

Note that this is a UI flow fix only: the model-side persistence path
(`Camera.set_calibration_result()` through `to_dict()`/`from_dict()`,
with `Machine._on_camera_changed` propagating to the config save) was
already correct and is unchanged.

### `LocalDeviceSource`

Wraps the previous USB and built-in camera behavior.

Responsibilities:

- discover local cameras
- open a device through OpenCV
- apply hardware-specific settings like resolution, white balance, contrast,
  brightness, and YUYV preference
- return frames to the controller

### `HttpSnapshotSource`

For cameras or endpoints that return one still image per request.

Behavior:

- performs a fresh HTTP/HTTPS request at most once every 2 seconds
- caches the last successfully decoded frame in memory for the current session
- if the endpoint temporarily returns non-image data or invalid image bytes,
  the cached frame is reused (for example, when the Creality Falcon A1 Pro
  stops returning valid images during a job run)
- if the endpoint goes offline or starts failing after a good frame was
  already captured, retries slow down to at most once every 10 seconds and
  warnings are throttled so the log does not get flooded
- decodes the returned image bytes with OpenCV
- returns a single frame

This is appropriate for “static image on each poll” endpoints.

### `HttpStreamSource`

For HTTP/HTTPS stream endpoints. In practice this is expected to cover common
network-camera stream URLs that OpenCV can read as a continuous source.

Behavior:

- opens the URL through OpenCV once
- keeps reading frames continuously

If the remote endpoint is an MJPEG-over-HTTP stream, this is the intended
source type to use.

### `RtspSource`

For RTSP camera feeds.

Behavior:

- opens the RTSP URL through OpenCV
- continuously reads frames

RTSP is intentionally modeled as its own source type even though the first
implementation is similar to the generic stream source. The reason is that it
is a materially different protocol and may need protocol-specific handling in
the future.

## Why not keep everything inside `CameraController`

That would have made the controller branch on every transport-specific detail:

- local camera probing
- local backend selection
- HTTP fetch behavior
- streaming URL handling
- future protocol-specific recovery logic
- settings support differences between local and network sources

By moving source acquisition behind a small interface, the controller remains
focused on camera logic that is genuinely shared.

## Migration strategy

The previous persisted camera shape stored `device_id` directly on the camera.

This PR migrates legacy data on load:

- old cameras without `id` get a generated UUID
- old `device_id` values are moved into `source_config.device_id`
- cameras without explicit `source_type` are treated as `local_device`
- Linux local-device migration still upgrades numeric IDs to persistent
  `/dev/v4l/by-id/...` paths where possible

This keeps existing camera setups usable while moving the data model to the new
shape.

## Manager and lookup changes

`CameraManager` used to key controllers by `device_id`.

It now keys them by stable camera `id`.

This is important because the following should not destroy the logical camera
identity:

- renaming a camera
- changing a USB path
- changing a URL
- replacing one source with another for the same logical camera slot

## GUI changes

The add-camera flow now starts with a source-type-first dialog.

### Rationale

The old flow assumed that “adding a camera” always meant selecting from a list
of discovered local devices. That does not scale to network sources because
they are often not discoverable in the same way.

The new flow first asks what kind of source the user wants to add, then shows
the relevant inputs for that type.

### Current add flow

1. Open **Add Camera**
2. Choose **Source Type**
3. Enter an optional **Name**
4. Depending on type:
   - **Local camera**: choose an attached device
   - **HTTP snapshot**: enter URL
   - **HTTP stream**: enter URL
   - **RTSP stream**: enter URL
5. Confirm to create the camera entry

This keeps local-device setup simple while making network sources possible in
the same UX entry point.

### Camera selection behavior

The camera settings page also now keeps the detail panel selection in sync with
the camera list more explicitly.

Behavior:

- when a new camera is added, that new camera becomes the selected one in the
  lower configuration panel
- when a selected camera is removed, selection moves to another remaining camera
  instead of leaving the lower panel focused on a stale entry
- when the last camera is removed, the lower panel is cleared

This matters because otherwise adding or removing multiple cameras can make it
look like the wrong camera is still active in the configuration area, which is
confusing during setup and testing.

### Local camera

Use this for:

- USB webcams
- built-in laptop cameras
- other cameras exposed as local capture devices

Configuration:

- choose the device from the local device list
- enable the camera
- optionally tune resolution, white balance, brightness, contrast, and YUYV

This source type supports hardware-style camera controls.

### HTTP snapshot

Use this when the URL returns one image per request, for example a camera
endpoint that serves a fresh JPEG or PNG whenever Rayforge polls it.

Configuration:

- paste the full `http://` or `https://` URL
- enable the camera

Expected behavior:

- Rayforge fetches a new image at most once every 2 seconds
- if the endpoint temporarily returns JSON or another non-image response,
  Rayforge keeps showing the last successfully received image from the current
  app session (for example the Creality Falcon A1 Pro stops returning valid images during a job run)
- if the endpoint later goes offline, Rayforge keeps showing the cached frame,
  retries more slowly, and avoids warning spam in the session log
- useful for low-frame-rate network cameras or simple web-exposed snapshots

### HTTP stream

Use this for continuous HTTP/HTTPS video streams.

Typical fit:

- MJPEG-over-HTTP feeds
- camera streams that OpenCV can open directly as a URL

Configuration:

- paste the full stream URL
- enable the camera

### RTSP

Use this for RTSP network cameras.

Configuration:

- paste the full RTSP URL, for example
  `rtsp://camera-host/stream`
- enable the camera

## Settings behavior differences

Not all settings apply equally to all source types.

Local-device hardware controls are meaningful for `local_device` sources:

- resolution
- auto/manual white balance
- brightness
- contrast
- YUYV preference

The add-camera local preview now avoids writing image controls back to the
device. It keeps preview capture non-destructive for brightness, contrast, and
white balance, then adopts the camera's current values when you actually add
that local camera to Rayforge.

Network sources generally do not expose those controls through this interface,
so the UI disables the hardware-specific options for them.

The following settings still remain meaningful across source types because they
operate on frames after acquisition:

- noise reduction
- transparency
- lens calibration
- image alignment

## What stays the same

This PR intentionally preserves the existing downstream camera behavior as
much as possible:

- preview rendering
- overlay on the work surface
- lens distortion correction
- image-to-world alignment
- homography-based work-surface transforms
- reconnect loop behavior

The major change is where frames come from, not how calibrated frames are used
once they are inside Rayforge.

### OpenCV is still the runtime backend for live streams

The first implementation of `http_stream` and `rtsp` uses OpenCV URL opening.
That keeps the refactor small and allows immediate support for common feeds,
but it means protocol support depends on the local OpenCV build.

Implication:

- some systems may handle HTTP streams or RTSP better than others
- if this needs improvement later, protocol-specific implementations can be
  added behind the same source abstraction without another model/controller
  redesign

### HTTP snapshot is fundamentally polling-based

This is correct for still-image endpoints, but it is not the same as a true
video stream. It may be lower frame rate and higher latency depending on the
server.

In this PR, snapshot polling is intentionally rate-limited to once every
2 seconds, and the last good frame is reused during temporary non-image or
decode failures.

### Local-device discovery remains local-only

The discovery list is still specifically for attached local cameras. Network
sources are added manually by URL.

This is intentional and keeps the UX honest.

## Network camera simulator

The repository includes a standalone test tool at
`scripts/camera/camera_simulator.py`. It generates a live clock image and
serves it through the three camera source types supported by this PR:

- HTTP snapshot: `http://127.0.0.1:5000/image.jpg`
- HTTP MJPEG stream: `http://127.0.0.1:5000/stream.mjpeg`
- RTSP stream: `rtsp://127.0.0.1:8554/live`

The HTTP endpoints are served directly by the script. The RTSP endpoint
requires a running [MediaMTX](https://github.com/bluenviron/mediamtx)
instance, started before the simulator. The simulator publishes the generated
frames to MediaMTX, which acts as the RTSP server.

Install the standalone tool dependencies in the Python environment used to
run it:

```bash
pip install av opencv-python numpy flask
```

Then start the simulator from the repository root:

```bash
./scripts/camera/camera_simulator.py
```

It prints the available URLs when it starts. In Rayforge, open the machine
settings, select the **Camera** tab, add a camera source matching the endpoint
you want to test, and enter the corresponding URL. The script has no runtime
options; its configuration is kept in the constants at the top of the file.

## Suggested manual/documentation ideas

If this becomes end-user documentation later, the following additions would be
useful:

- examples of valid snapshot, MJPEG, and RTSP URLs
- troubleshooting by source type
- note that stream support depends on OpenCV/media backend availability
- note which settings only apply to local devices
- recommended camera placement and alignment workflow

## Files and areas touched

Main refactor areas:

- `rayforge/camera/models/camera.py`
- `rayforge/camera/source.py`
- `rayforge/camera/controller.py`
- `rayforge/camera/manager.py`
- `rayforge/camera/v4l.py`

UI integration:

- `rayforge/ui_gtk/camera/selection_dialog.py`
- `rayforge/ui_gtk/camera/camera_preferences_page.py`
- `rayforge/ui_gtk/camera/properties_widget.py`
- `rayforge/ui_gtk/camera/image_settings_widget.py`
- `rayforge/ui_gtk/camera/display_widget.py`
- `rayforge/ui_gtk/camera/alignment_widget.py`
- `rayforge/ui_gtk/camera/capture_surface.py`
- `rayforge/ui_gtk/camera/lens_calibration_widget.py`
- `rayforge/ui_gtk/camera/wizard/wizard.py`
- `rayforge/ui_gtk/camera/wizard/base_page.py`
- `rayforge/ui_gtk/camera/wizard/image_settings_page.py`
- `rayforge/ui_gtk/camera/wizard/capture_page.py`
- `rayforge/ui_gtk/camera/wizard/alignment_page.py`
- `rayforge/ui_gtk/camera/wizard/lens_calibration_settings_page.py`
- `rayforge/ui_gtk/machine/settings_dialog.py`

Camera consumers updated to use stable camera IDs:

- `rayforge/ui_gtk/canvas2d/surface.py`
- `rayforge/builtin_addons/rayforge-addon-sketcher/sketcher/ui_gtk/sketchcanvas.py`

Tests:

- `tests/camera/test_camera.py`
- `tests/camera/test_camera_manager.py`
- `tests/camera/test_controller.py`
- `tests/camera/test_source.py`
- `tests/camera/test_v4l.py`
- `tests/ui_gtk/camera/test_properties_widget.py`
- `tests/ui_gtk/camera/test_selection_dialog.py`
- `tests/ui_gtk/camera/test_wizard_subscriptions.py`
- `tests/ui_gtk/camera/test_wizard_calibration_flow.py`

## Summary

This PR changes the camera system from a local-device-only model to a
source-based model while preserving the existing image-processing and alignment
pipeline.

The key architectural decisions are:

1. stable UUID camera identity
2. explicit source type + source config
3. acquisition abstraction below `CameraController`
4. source-type-first add-camera UI
5. serialized source switching and cancellation-aware capture lifecycle,
   incorporating architectural ideas from PR #435
6. synchronized frame publication and non-blocking device discovery
7. balanced, idempotent camera stream subscriptions across the camera UI
8. a wizard flow where lens calibration is saved and followed by
   alignment instead of ending the wizard

Those choices were made to support more camera transports now without making
future camera sources harder to add later.


rayforge/camera/source.py
```diff
@@ -0,0 +1,671 @@
...
+    def _log_warning(self, now: float, message: str, *args) -> None:
+        if self._should_log_warning(now):
+            logger.warning(message, *args)
+
+    def read_frame(self) -> np.ndarray | None:
+        now = time.monotonic()
+        if self._last_poll_time is not None:
+            elapsed = now - self._last_poll_time
+            if elapsed < self._next_poll_interval:
+                return self._cached_frame()
```
MausRundung: This Is not good practice... since this will loop and crash at 100% utilization.

When it is working, returning the cached frame on every spin between the 2-second polls means we're doing a memcpy via .copy() and pushing duplicate frames through the whole undistort + GTK redraw pipeline at 30–60 FPS for a static image.

rayforge/camera/source.py
```diff
@@ -0,0 +1,671 @@
...
+        finally:
+            self.cap = None
+
+    def read_frame(self) -> np.ndarray | None:
+        if self.cap is None:
+            raise OSError("Camera source is not open")
+        ret, frame = self.cap.read()
+        if not ret or frame is None:
+            return None
+        return frame
```
MausRundung, line 523: Notice how LocalDeviceSource tracks self._read_failures and raises an OSError after MAX_CONSECUTIVE_READ_FAILURES? We need that here too.

If an RTSP or HTTP camera drops Wi-Fi or reboots, self.cap.read() just returns (False, None) on every call. Because we never raise OSError, the controller never triggers a reconnect, and reconnect_delay_seconds = 10.0 is essentially dead code. The thread will just loop indefinitely pulling None.

rayforge/camera/source.py
```diff
@@ -0,0 +1,671 @@
...
+            elapsed = now - self._last_poll_time
+            if elapsed < self._next_poll_interval:
+                return self._cached_frame()
+
+        request = urllib.request.Request(
+            self._uri,
+            headers={"User-Agent": "Rayforge Camera"},
+        )
+        try:
+            with urllib.request.urlopen(
```
MausRundung, line 644: Needs a safety limit on response.read(). Users frequently mix up "HTTP Stream" (MJPEG) and "HTTP Snapshot". If someone pastes an MJPEG stream URL into the snapshot field, read() will buffer an infinite chunked stream straight into RAM until the OOM killer nukes the process.

A sanity cap like response.read(20 * 1024 * 1024) (20MB) should be plenty for any still frame and protect against runaway memory.

rayforge/camera/source.py
```diff
@@ -0,0 +1,671 @@
...
+
+    def read_current_settings(self) -> dict[str, object]:
+        if self.cap is None:
+            raise OSError("Camera source is not open")
+        cap = self.cap
+        settings: dict[str, object] = {
+            "contrast": float(cap.get(cv2.CAP_PROP_CONTRAST)),
+            "brightness": float(cap.get(cv2.CAP_PROP_BRIGHTNESS)),
+        }
+        auto_wb = cap.get(cv2.CAP_PROP_AUTO_WB)
```
MausRundung, line 464: Cams where the diver will not support White Balance should drop this

rayforge/ui_gtk/machine/settings_dialog.py
```diff
@@ -392,19 +394,67 @@ def _on_show_toast(self, sender, message: str):
...
+            source_config=source_config,
         )
+        if new_camera.source_type is CameraSourceType.LOCAL_DEVICE:
+            self._adopt_current_local_camera_settings(new_camera)
         new_camera.enabled = True
+        self.camera_page.select_camera_by_id(new_camera.id)
         self.machine.add_camera(new_camera)
         # The machine.changed signal will handle the UI update
 
+    def _adopt_current_local_camera_settings(self, camera: Camera) -> None:
```
MausRundung, line 436: Add something that will drop the Controller once its no longer needed otherwise another application might run into problems like obs or discord

MausRundung (changes requested):
A few things that should be fixed here but in general looks good

MausRundung:
Ah and run format, and lint there also seem to be a few issues there

rayforge/camera/source.py
atkaper: Fixed in ef61655b9. HttpSnapshotSource now blocks the capture thread until the next snapshot poll is due instead of returning the cached frame on every capture-loop iteration. Cached-frame reuse is still preserved for failed/offline fetches, but it now happens at the configured poll/retry interval rather than 30–60 FPS, so we avoid repeated memcpy/processing/redraw work for the same snapshot.

rayforge/camera/source.py
atkaper: Fixed in f2d0470df. OpenCvUrlSource now mirrors LocalDeviceSource’s consecutive read-failure handling: RTSP/HTTP streams count failed cap.read() calls, reset the counter after a good frame, and raise OSError after repeated failures. That lets the controller hit its reconnect path and use reconnect_delay_seconds instead of continuing to read None indefinitely.

rayforge/camera/source.py
atkaper: Fixed in 1f5c0a1ac. HTTP snapshot reads are now capped at 20MB + 1 byte so a misconfigured MJPEG/stream URL cannot be buffered indefinitely. Oversized responses are rejected before decoding, the source backs off to the offline retry interval, and the last good frame is reused when available.

MausRundung:
i did not check does auto upscaling on camera images still work and does noise reduction still work?

rayforge/camera/source.py
atkaper: Fixed in 71dd8d503. LocalDeviceSource now only reports a manual white-balance value when the driver returns a plausible temperature. Cameras/drivers that expose unsupported white-balance controls as zero now omit white_balance instead of adopting a fake clamped setting.

rayforge/ui_gtk/machine/settings_dialog.py
atkaper: Fixed in b7dc674f1 and ba76d7de3. The temporary CameraController used to adopt current local-camera settings is now disposed in a finally block after reading settings, including the error path, so it releases resources and disconnects model signal handlers once it is no longer needed. Added tests for both success and failure paths.

atkaper:
@MausRundung 

> i did not check does auto upscaling on camera images still work and does noise reduction still work?

Same libraries are used, so both should still be there. I just tested noise reduction by setting it to 100%, and works fine.
Not sure how to manually test upscaling.

Note: I did process all your review comments, and fixed the build (one test was failing in github runner, while it did pass on my local machine. Only once I ran it locally in a docker image, it was reproducible. It was a test case where it was altering available video sources, and on my local run my real linux video sources were just taken in the test).

MausRundung:
if you ask me looks good for merging ... but @knipknap will probably take a look him self once he is back

rayforge/camera/controller.py
```diff
@@ -11,13 +9,28 @@
...
 
 if TYPE_CHECKING:
     from gi.repository import GdkPixbuf
 
+    from .source import CameraSource
+
 logger = logging.getLogger(__name__)
 
+
+def __getattr__(name: str):
```
knipknap, line 26: This module level getattr block worries me, as does string matching on module names and imports a private function. This whole block points to a layer violation.

rayforge/ui_gtk/camera/lens_calibration_widget.py
```diff
@@ -92,6 +92,9 @@ def _on_distortion_value_changed(
             return
         setattr(self.camera, config_key, spin_row.get_value())
 
+    def start(self) -> None:
+        self.camera.settings_changed.connect(self._on_camera_settings_changed)
```
knipknap, line 96: This already exists also in the constructor - double connection risk?

website/docs/machine/camera.md
knipknap, line 1: Translated versions of this page should also be updated. AI usually does this well.

rayforge/ui_gtk/camera/selection_dialog.py
```diff
@@ -1,265 +1,344 @@
...
 
+_SCANNING_PLACEHOLDER = _("Scanning for devices\u2026")
+
 
 class CameraSelectionDialog(Adw.MessageDialog):
     def __init__(
         self,
         parent,
-        mode: Literal["available", "configured"] = "available",
+        active_controllers: list[CameraController] | None = None,
```
knipknap: Removing the mode breaks the smart-stock addon. The stock detection addon needs the "configured" mode, as the dialog is used to select one of the already configured cameras. For reference, the addon is here: https://github.com/barebaric/rayforge-addon-smart-stock

rayforge/camera/source.py
knipknap, line 1: Could you split this into multiple files in a module, i.e. one `source/base.py` and additional files per source?

rayforge/camera/source.py
```diff
@@ -0,0 +1,708 @@
...
+    MIN_POLL_INTERVAL_SECONDS = 2.0
+    OFFLINE_RETRY_INTERVAL_SECONDS = 10.0
+    WARNING_LOG_INTERVAL_SECONDS = 30.0
+    MAX_SNAPSHOT_BYTES = 20 * 1024 * 1024
+
+    def __init__(self, config: Camera):
+        super().__init__(config)
+        # Owned/frozen at construction time; see LocalDeviceSource.__init__.
+        self._uri: str = config.source_uri
+        self._timeout = float(config.source_config.get("timeout_seconds", 5.0))
```
knipknap, line 573: Instead of relying on non type safe string matching, I think source should have a to_dict() and from_dict(), and Source subclasses could use typed attributes.
For forward compatibility, app convention is to also have an "extra" attribute that is used as a bin for unknown attributes, in case a user happens to open an older version of the app in a project that has newer config.

rayforge/ui_gtk/camera/selection_dialog.py
knipknap, line 1: Some issues I noticed when in the new camera selector:
- Removing the mode breaks usage as a plain selection dialog, as needed by the smart-stock addon. The stock detection addon needs the "configured" mode, as the dialog is used to select one of the already configured cameras. For reference, the addon is here: https://github.com/barebaric/rayforge-addon-smart-stock . In general, I the intent of the dialog was to separate selection from configuration - giving it a name field mixes these concerns. Honestly I would remove the field, or at least fill it by default with the detected name.
- On my machine the dialog changes height, depending on drop down / camera selection. This makes it feel a bit "jumpy" during use
- Perhaps the "source type" could be set to "local" by default and a first device could be selected? That would allow for some image to show immediately like in the old selector.

rayforge/ui_gtk/camera/properties_widget.py
```diff
@@ -19,16 +21,49 @@ def __init__(self, controller: CameraController | None, **kwargs):
...
+        self.source_entry.set_width_chars(40)
+        self.source_entry.set_hexpand(True)
+        self.source_entry.connect("changed", self._on_source_changed)
+        self.source_entry.connect("activate", self._commit_source_uri)
+        focus_controller = Gtk.EventControllerFocus()
+        focus_controller.connect("leave", self._commit_source_uri)
+        self.source_entry.add_controller(focus_controller)
+        self.source_row.add_suffix(self.source_entry)
+
+        self.source_combo = Gtk.DropDown(model=Gtk.StringList.new([]))
```
knipknap, line 58: This should use Adw.ComboRow. Note that there are some gotchas using the Adw.ComboRow, probably best adopt the patterns used on other preferences pages, such as the General Preferences page Driver selector, or the Panel orientation drop down in the Hardware settings page.

rayforge/ui_gtk/camera/selection_dialog.py
```diff
@@ -1,265 +1,344 @@
...
+        self.uri_row.set_visible(
+            source_type
+            in {
+                CameraSourceType.HTTP_SNAPSHOT,
+                CameraSourceType.HTTP_STREAM,
+                CameraSourceType.RTSP,
+            }
+        )
+        if not is_local:
+            self.preview_image.set_pixbuf(None)
```
knipknap, line 170: This creates a warning:
tests/ui_gtk/camera/test_selection_dialog.py::test_dialog_opens_instantly_with_scanning_placeholder
  /home/sam/Documents/projects/code/rayforge/rayforge/ui_gtk/camera/selection_dialog.py:170: DeprecationWarning: Gtk.Picture.set_pixbuf is deprecated
    self.preview_image.set_pixbuf(None)

knipknap (changes requested):
Thanks Thijs — this is a lot of careful work.
Also Thanks to MausRundung for the reviews!

I've left detailed comments inline. None of them are conceptually hard, but the camera-selector regression matters because the smart-stock addon depends on it, so I'd like that and the typed source-config cleanup addressed before merge.

Happy to re-review once they're in.

atkaper:
@knipknap Thanks for the feedback! I'll have a look at it this evening or tomorrow evening.

rayforge/camera/controller.py
atkaper: Addressed in commit c6576377bb4b5fbfb7fb6934feec010a3e6cfe9d. Removed the module-level getattr compatibility shim and updated the regression test to import to_videocapture_arg from its owning camera.source module. This removes the dynamic module API, private re-export, and layer violation.

rayforge/ui_gtk/camera/lens_calibration_widget.py
atkaper: Addressed in commit 581d2a4bb89c741e843109ada508c0cbbd953d00. The lens calibration widget now guards its camera-settings signal connection, making start() idempotent. This prevents the constructor connection and lifecycle start() call from registering the handler twice, while preserving proper disconnect/reconnect behavior when the wizard page is left and re-entered.

website/docs/machine/camera.md
atkaper: Addressed in commits fd832bf1d5ba9633ff88784567f081c28c32dd9b and 417157f13c945238e7bceb25ef99b8b3a80ad713.
Updated all six translated camera documentation pages with localized guidance for network camera types, adding network cameras, updating URLs without losing calibration/alignment, and troubleshooting connection issues. Also aligned the translated setup sections with the source documentation by removing obsolete USB-only wording, outdated navigation instructions, toolbar-button guidance, and legacy device-index examples. The current Machine Settings flow and local camera source selection are now documented consistently.

rayforge/camera/source.py
atkaper: Addressed in commit f74be5c578fb2e377d1ca08d812ee81fd563a1e1. Split the 709-line camera source module into a camera/source/ package:
• base.py — shared source abstractions, validation, and capture helpers
• local.py — local device capture and discovery
• network.py — RTSP, HTTP stream, and HTTP snapshot sources
• factory.py — source construction and local-device ID listing
• init.py — public source API
Existing imports and behavior remain compatible, and the targeted camera tests and project lint/type checks pass.

rayforge/camera/source.py
atkaper: Addressed in commit 3e56e66e9f92cb2a56a3101ef2cf9d74ebd02b6c. Replaced string-key access for source configuration with typed source-config dataclasses and source-level to_dict()/from_dict() support:
• Local sources use a typed LocalDeviceConfig
• HTTP/RTSP streams use typed URL configuration
• HTTP snapshots use typed URL and timeout configuration
• Unknown source attributes are retained in each config’s extra dictionary for forward compatibility
• Camera serialization now delegates source-config normalization through the source factory
• Legacy device_id data remains supported
Added a regression test confirming unknown source configuration attributes survive a serialize/deserialize round trip.

rayforge/ui_gtk/camera/selection_dialog.py
atkaper: Addressed in 7fc0dce0789753955afe6ebe0755e8ababb95df6.
• Restored mode="configured" for plain selection of existing cameras, returning the configured camera model ID through selected_device_id for the smart-stock addon.
• Removed the name field so selection and camera configuration remain separate concerns. New cameras now derive their initial name from the device or URL.
• Stabilized the dialog contents with a fixed minimum content height.
• New-camera mode now defaults to local cameras, selects the first discovered device automatically, and starts its preview.
• Added UI tests covering configured selection, name-field removal, local defaults, and automatic first-device selection.

MausRundung:
once your done can you merge all the commits into one so its easier to review

rayforge/ui_gtk/camera/properties_widget.py
atkaper: Addressed in 61d09a4fd5e1761eade263fa669bcde90ebd84db.
The local camera selector now uses Adw.ComboRow, following the established preferences-page patterns. URL editing remains in a separate Adw.ActionRow for network sources, and the existing device selection and scan safeguards are preserved. Added a regression assertion confirming the selector type.

rayforge/ui_gtk/camera/selection_dialog.py
atkaper: Addressed in d6fc8c76bb1eb4fcd682223fa02726a4f21c121f.

Replaced all deprecated Gtk.Picture.set_pixbuf() calls with set_paintable(), converting preview pixbufs through Gdk.Texture.new_for_pixbuf(). The camera selection UI tests now pass without the deprecation warning.

atkaper:
@MausRundung 

> once your done can you merge all the commits into one so its easier to review

I merged them all into one commit, from the review comments of @knipknap onwards. I hope that was what you meant? Or do you mean the whole PR?

rayforge/camera/source/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+
+__all__ = [
+    "CameraSource",
+    "HttpSnapshotSource",
+    "HttpStreamSource",
+    "LocalDeviceSource",
+    "OpenCvUrlSource",
+    "RtspSource",
+    "SourceDescriptor",
+    "_capture_has_initial_frame",
```
knipknap: Private (= underscore prefixed) functions should not be exposed.

rayforge/camera/source/base.py
```diff
@@ -0,0 +1,220 @@
...
+import time
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+from typing import Any, ClassVar
+from urllib.parse import urlsplit
+
+import cv2
+import numpy as np
+
+from ..models.camera import Camera, CameraSourceType
```
knipknap: Camera is only used for typing here, so should be in a TYPE_CHECKING block. This also resolves the circular import that  led to using a dynamic/local import in camera.py

rayforge/camera/models/camera.py
```diff
@@ -647,12 +647,16 @@ def source_display_value(self) -> str:
         return self.source_uri
 
     def to_dict(self) -> dict[str, Any]:
+        from ..source import source_config_to_dict
```
knipknap: This should be at the top of the file.

rayforge/camera/source/local.py
```diff
@@ -0,0 +1,318 @@
...
+
+import cv2
+import numpy as np
+
+from ..models.camera import Camera
+from .base import (
+    CameraSource,
+    SourceConfig,
+    SourceDescriptor,
+    _capture_has_initial_frame,
```
knipknap: Private (=underscore prefixed) functions should never be imported

rayforge/ui_gtk/camera/selection_dialog.py
```diff
@@ -146,6 +147,29 @@ def _build_ui(self) -> None:
         self.uri_row = uri_row
         source_group.add(uri_row)
         self._update_visibility()
+        self.type_row.set_selected(1)
+        self.device_row.set_visible(True)
+        self.preview_group.set_visible(True)
+
+    def _load_configured_cameras(self) -> None:
+        from ...context import get_context
```
knipknap: This should be at the top of the file

tests/camera/test_controller.py
```diff
@@ -171,7 +171,7 @@ def test_get_work_surface_image_no_image_data():
 
 
 def test_to_videocapture_arg():
-    from rayforge.camera.controller import _to_videocapture_arg
+    from rayforge.camera.source import _to_videocapture_arg
```
knipknap: this should be at the top of the file

atkaper:
@knipknap can you re-run the job which failed on some timeout? Otherwise I need to do some dummy commit:
```gh run rerun 35904581964 --repo barebaric/rayforge --failed```
Or from web: → https://github.com/barebaric/rayforge/actions/runs/35904581964
Thanks!

knipknap (changes requested):
Thanks again. I found only a few minor things now. Otherwise looks ready! I did not test the actual network camera, as I have none, so I only checked for regressions - both my local cameras still work!

atkaper:
@knipknap I will have a look at the new findings. And you can simply test the network camera's by running the scripts/camera/camera_simulator.py. If you start it without mediamtx you should still have the two http sources. If you run it with mediamtx, you can also test the rtsp stream.

MausRundung:
one more thing why do you sometimes use self._device_id and sometimes self._source_config["device_id"]

rayforge/camera/source/__init__.py
atkaper: Handled in e24e12649e8c2d79908bac68774e6af57b1939a0

rayforge/camera/source/base.py
atkaper: Handled in e24e12649e8c2d79908bac68774e6af57b1939a0

rayforge/camera/models/camera.py
atkaper: Handled in e24e12649e8c2d79908bac68774e6af57b1939a0

rayforge/camera/source/local.py
atkaper: Handled in e24e12649e8c2d79908bac68774e6af57b1939a0

rayforge/ui_gtk/camera/selection_dialog.py
atkaper: Handled in e24e12649e8c2d79908bac68774e6af57b1939a0

tests/camera/test_controller.py
atkaper: Handled in e24e12649e8c2d79908bac68774e6af57b1939a0

atkaper:
@MausRundung 

> one more thing why do you sometimes use self._device_id and sometimes self._source_config["device_id"]

Handled in e24e12649e8c2d79908bac68774e6af57b1939a0

atkaper:
@knipknap / @MausRundung Ok, I tried handling all comments. I hope I did not miss any ;-) It all became much bigger than anticipated. Oh well...

MausRundung:
no worries its always like this on such core changes...

knipknap (approved):
Thumbs up from me, looks good!

MausRundung (approved):
Looks good happy you went though with all our Requests :)

knipknap:
Thank you for patiently resolving all issues!

